PR #196, opened 2017-02-09
honours user-specified install prefix

ScarabMonkey:
Minor changes to cmake scripts that honour the CMAKE_INSTALL_PREFIX which the user can send to cmake.
This allows shared fftw library to be be built and installed to the correct location with 'make install' - rather than assuming the original source directory is available across the cluster (for example).


bforsbe:
v2.0.4 incorporates these changes. Thanks for pointing it out!

PR #218, opened 2017-03-22
Use markdown formatting

smsaladi:
Not sure if the formatting was on purpose, but making these modifications will tell Git how to display the headings and code (just a little bit prettier)


biochem-fan:
Thank you very much for your suggestion. Part of this patch will be included in the next release (2.1).
We don't include the 'shell' tag because it might be confusing to people viewing the file without Markdown formatting.

PR #233, opened 2017-05-25
fixed installation instructions and markdown formatting

jewettaij:
A member of our lab was unable to follow the installation instructions in the README.md, and I took this opportunity to change the README.md file in the places where he was stuck. (Specifically, you need to install several other packages besides cmake and git, specifically the packages which include: g++, mpi, and fftw) These changes were tested with ubuntu16.04


biochem-fan:
Thank you very much for your suggestion. This patch will be included in the next release (2.1).

PR #261, opened 2017-07-18
added missing write(DO_LOCK)

andschenk:
When trying to mark a not yet started job as finished, relion doesn't release the lock file for the default_pipeline.star. I added a write(DO_LOCK) before the return statement in order to release the lock in this case.


scheres:
This looks OK.

PR #269, opened 2017-08-08
Show total number of selected particles in "show metadata this class"

biochem-fan:
and correctly escape @ marks.

See the example below:
image


biochem-fan:
Merged to the developer branch after bug fix.

PR #290, opened 2017-09-29
Don't link non-GUI apps against X11

milliams:
When compiling Relion with the default `DGUI=ON`, it links all of the applications in the suite against X11, even apps like `relion_refine` which do not depend on any X11 features. This causes an issue on some batch clusters/supercomputers as X11 is often not present on compute nodes.

This PR introduces a new convenience library, `relion_gui_lib` which contains only the GUI parts and removes them from `relion_lib`.

This allows the relion GUI application to link against X11 etc. while, e.g., relion_refine to be free of X11 dependencies.

It also increases the minimum version of CMake to 3.3 as it uses the [`if(... IN_LIST ...)`](https://cmake.org/cmake/help/v3.8/command/if.html) syntax.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,16 +1,7 @@
-cmake_minimum_required(VERSION 2.8.12)
-set(RELION_CMAKE_MINIMUM_REQUIRED_VERSION "2.8.12")
+cmake_minimum_required(VERSION 3.3)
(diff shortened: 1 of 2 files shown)


bforsbe:
Looks good but I'll need to try it out to see that it works more generally before I sign off on this. Good idea and execution by the looks of it though. Thanks!

biochem-fan:
Your suggestion has been incorporated in 2.1 (8228bbd). Thank you very much.

PR #338, opened 2018-02-02
Expose the queue name and queue submit command as environment variables.

JanneM:
Make the submission queue name; submission command; max MPI processes; max threads and whether to run locally or as a submission script settable by environment variables instead of hard-coding them. 

These are all specific to the particular system (many cluster installations only accept job submissions). With these changes we can set them to sensible defaults for our users instead of asking them to change it themselves (and them not reading the documentation, then wonder why things don't work). This helps us provide a smoother experience for our users, with fewer errors.


jmansour:
Came here to submit a pull request for just the queue command, but @JanneM you've beaten me to it. 

Might I suggest you also add the environment variable flags to the help text associated with each option? 🙂

JanneM:
@jmansour Good idea. Added a line for each new environment variable. I also added "yes" and "no" as valid parameters for boolean inputs. That makes the environment setting the same as the visual parameter in the UI, so it should improve usability a little bit.

biochem-fan:
Thank you very much for your suggestion. This will be included in the next version (3.0).

PR #366, opened 2018-06-12
Fixed overflow bug caused by relion_MPI_Bcast

lqhuang:
(1) Modify type of `count` arguments from `int` to `long int`/`std::ptrdiff_t` for relion_MPI_Bcast().
(2) Reimplement `relion_MPI_Bcast()`. If `count` is smaller than a certain size (1 GB), broadcast data to each slave directly. If `count` is larger than this threshold, data is sent to each slave by `relion_MPI_send()` and `relion_MPI_Recv()`.

### DIFF
--- a/src/ml_optimiser_mpi.h
+++ b/src/ml_optimiser_mpi.h
@@ -23,15 +23,7 @@
 #include "src/mpi.h"
 #include "src/ml_optimiser.h"
 
-#define MPITAG_JOB_REQUEST 0
-#define MPITAG_JOB_REPLY 1
-#define MPITAG_METADATA 2
-#define MPITAG_RANDOMSEED 3
-#define MPITAG_IMAGE 4
-#define MPITAG_PACK 5
-#define MPITAG_RFLOAT 6
-#define MPITAG_INT 7
-#define MPITAG_IDENTIFIER 8
+// definition of MPITAG has been moved to header mpi.h
 
 class MlOptimiserMpi: public MlOptimiser
 {
--- a/src/mpi.cpp
+++ b/src/mpi.cpp
@@ -254,15 +254,40 @@ int MpiNode::relion_MPI_Recv(void *buf, std::ptrdiff_t count, MPI_Datatype datat
 }
 
 
-int MpiNode::relion_MPI_Bcast(void *buffer, int count, MPI_Datatype datatype, int root, MPI_Comm comm)
+int MpiNode::relion_MPI_Bcast(void *buffer, long int count, MPI_Datatype datatype, int root, MPI_Comm comm)
 {
 	int result;
+    int unitsize(0);
+    MPI_Type_size(datatype, &unitsize);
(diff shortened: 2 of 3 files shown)


lqhuang:
In several cases which the box size of virus particle is very large (> 1000 pixel), `relion_refine_mpi` program raised MPI_error and exit.

Error traceback looks like this:

    0: Other MPI error, error stack:
    PMPI_Bcast(2231)........: MPI_Bcast(buf=0x7f17b405e010, count=941192000, MPI_DOUBLE, root=1, 
    MPI_COMM_WORLD) failed
    MPIR_Bcast_impl(1798)...: 
    MPIR_Bcast(1826)........: 

After debugging, I found it caused by integer overflow problem of `relion_MPI_Bcast`. Previous codes used MPI_Bcast() function to broadcast data to other slaves.  But sometimes `count` argument is larger than max `int`, eg:

line 770 in `ml_optimiser_mpi.cpp`:

    node->relion_MPI_Bcast(MULTIDIM_ARRAY(mymodel.PPref[i].data),
        MULTIDIM_SIZE(mymodel.PPref[0].data), MY_MPI_COMPLEX, sender, node->slaveC);

line 2242 in `ml_optimiser_mpi`:

    node->relion_MPI_Bcast(MULTIDIM_ARRAY(mymodel.Iref[ith_recons]),
        MULTIDIM_SIZE(mymodel.Iref[ith_recons]), MY_MPI_DOUBLE, reconstruct_rank, MPI_COMM_WORLD);

and more.

The problem is `MULTIDIM_SIZE` returns a `long int` number and relion_MPI_Bcast() only accept `count` in `int` type.

    #define MULTIDIM_SIZE(v) ((v).nzyxdim)  // line 129 in multidim_array.h
    // Number of elements in NZYX
    long int nzyxdim; // line 477 in multidim_array.h

Some codes probably should use `relion_MPI_Send()` and `relion_MPI_Recv()` instead of `relion_MPI_Bcast()`. Everything works well with small boxsize of particle. However, if the particle size is really large (>1000 pixel), the variable `count` overflowed and would be negative when passing into `relion_MPI_Bcast()`. And MPI_BCast() is also limited under a maximum aount (2 GB) of data can be sent. Hence, a broadcast error raised in this situation.

What I did to fix this bug:
(1) Modify type of `count` arguments from `int` to `long int`/`std::ptrdiff_t` for relion_MPI_Bcast().
(2) Reimplement relion_MPI_Bcast(). If `count` is smaller than a certain size (1 GB), broadcast data to each slave directly. If `count` is larger than this threshold, data is sent to each slave by `relion_MPI_send()` and `relion_MPI_Recv()`.

It's a very safe and compatible way to fix this problem. We have successfully compiled and run it to refine our virus particle.

Regards,
Lanqing

biochem-fan:
Thank you very much for your suggestion. We will test this.

biochem-fan:
This will be included in the next release (3.0) of RELION. Thank you very much.

Issue #453, opened 2019-03-12
relion 3 compiled sucessfully with CUDA or ALTCPU but no GUI and no error message 


riverseow:
I am trying to get the latest relion working on a workstation with Ubuntu 18.10 and Geforce 1080 Ti graphic card. I had an old Relion 1.3 works fine. Now when I compiled the latest Relion 3.0. I first tried CUDA 10.1. Everything seems fine but after compile and install, when I execute relion, no GUI display, one relion process using 100% of the CPU. No crashes, after answer "y" to the question about start a new project here, nothing display. I thought maybe it is CUDA. I tested CUDA compiler using CUDA's samples and everything runs fine.  Then I did a new compilation of Relion 3.0 with -DALTCPU under a new directory. Everything looks fine and again, after typing relion, no GUI, no error message. 

Any hints to solve the problem will be appreciated. Should I downgrade to CUDA 9 and install GCC 6.0. I installed CUDA10.1 because using previous CUDA, it complains that current GCC 8.2 cannot compile the CUDA. 

River

bforsbe:
Are you launching relion as a background process?

riverseow:
No. I just type "relion" and enter, no "&" after the command. In previous version, this should bring up the GUI but I see nothing but only 100% for 1 cpu is running for ever. Some lock file and default_pipeline.star file were written but no other messages or anything. I don't have anyway to find out what is the problem. Is there any debug mode to print to standout for relion to let me find out what step got hold for ever.

bforsbe:
if you compile using cmake flags 

´´´ -DCMAKE_BUILD_TYPE=DEBUG -DBUILD_SHARED_LIBS=OFF ´´´

you can run relion through gdb and find out where it is catching. Using your debug-binary, run 

´´´ gdb --args relion ´´´

and hit 'r' to run it. If it seems stalled, stop it through ctrl+c and ask for a backtrace by typing 'bt' and hit enter.

riverseow:
Bjorn,

Thank you so much. I followed your instructions. First, I got an error
message "cannot open display on localhost:0.0". Instead of running directly
from the ubuntu terminal. I started an Xterm which make sure to start X
Windows OK and now it seems OK. I got the GUI. No more other errors.  I
thought that was the problem. However, when I compiled again without those
debug modes, I still cannot get the GUI with xterm. I thought maybe the
BUILD_SHARED_LIB=OFF is enough. I compiled with that flag only, still
[...]

PR #463, opened 2019-05-03
Fix a couple of minor bugs.

akesandgren:


biochem-fan:
Thank you very much, but we have already fixed this for https://github.com/3dem/relion/issues/453 and been testing it internally. I am sorry but I will reject this pull request. Our fix will be public next week.

PR #469, opened 2019-05-16
add threads option to postprocess automask

fuzikt:
Adding threading option available in relion_mask_create to relion_postprocess (automask).


biochem-fan:
Are you using `--auto_mask`? Honestly speaking, I am considering deprecating this option. This feature is not available from GUI already.

fuzikt:
Not using gui. Maybe too oldfashioned script fan.... (originating form relion 1.4).

biochem-fan:
I discussed with Sjors and decided to remove this option from RELION 3.1, because many other features (CtfRefine, Polish, etc) now require PostProcess to have been done with a proper mask. Users should make a mask as a separate job, inspect if it is good and soft enough, and proceed.

PR #478, opened 2019-05-28
Register build byproducts for FFTW external.

bbockelm:
If an external project is used for FFTW / FFTWF, Relion links libraries directly against the resulting libraries (such as `libfftw3f.so`).

Some build generators (specifically, Ninja) require that the full dependency graph is calculated at the beginning of the build process.  Since the file `libfftw3f.so` does not exist at the beginning of the build (the `ExternalProject` command has not been run yet) -- and there is no rule to generate that file -- the build fails.

With this change, we introduce a fake target that relies on the `ExternalProject` and has as a build byproduct the resulting library. This allows `ninja` to correctly resolve the dependency and build Relion.

### DIFF
--- a/cmake/BuildFFTW.cmake
+++ b/cmake/BuildFFTW.cmake
@@ -85,6 +85,13 @@ else()
 	#	LOG_BUILD
 	LOG_INSTALL)
 	
+
+	add_custom_command(
+		COMMAND ${CMAKE_COMMAND} -E echo "Registering own FFTW byproducts"
+		OUTPUT "${FFTW_EXTERNAL_PATH}/lib/libfftw3.so"
+		DEPENDS own_fftw_lib)
+	add_custom_target(own_fftw_lib_byproducts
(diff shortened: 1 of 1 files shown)


biochem-fan:
Thank you very much for your contribution. We will merge this to version 3.1 (under active development; public beta testing hopefully in this autumn) so that we have enough time for testing.

bbockelm:
Sounds great!

I should mention that, to reproduce the current build failure, you can run:

```
cmake .. -GNinja
```
(assuming `..` is the source directory and the `ninja` generator is installed).   Fails without this patch; fix with it.  Should be (effectively) a no-op for the `make` generator.

biochem-fan:
Thanks for instructions. I merged this to our internal development branch for 3.1 and confirmed Ninja worked fine. So I close this PR.

FYI: On my old cmake 2.8.12.2, I had to rename `ninja` binary to `ninja-build`.
https://stackoverflow.com/questions/38658014/ninja-not-found-by-cmake

PR #546, opened 2019-11-22
Fix typo in tiff_converter

js947:
currently preventing ver3.1 from building with TIFF support


biochem-fan:
Thanks! Interestingly, the old code compiled on GCC 4.8.5.

PR #550, opened 2019-11-25
Add environment variable for default scratch directory.

tjragan:
This is a simple addition to the gui to pickup the environment variable RELION_SCRATCH_DIR and use it for all 5 jobs that use scratch directories.


biochem-fan:
I think this makes sense. Thanks for contribution.

PR #575, opened 2020-02-12
OPTIM: merge data transfers. part 1

arom4github:
some data transfers are merged into a single one

### DIFF
--- a/src/acc/acc_ml_optimiser_impl.h
+++ b/src/acc/acc_ml_optimiser_impl.h
@@ -2314,22 +2314,18 @@ void storeWeightedSums(OptimisationParamters &op, SamplingParameters &sp,
 
 		DEBUG_HANDLE_ERROR(cudaStreamSynchronize(cudaStreamPerThread));
 
+		// here we introduce offsets for the clases in an array as it is more efficient to
+		// copy one big array to/from GPU rather than four small arrays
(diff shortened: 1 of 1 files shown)


biochem-fan:
Thanks for contribution. What does "part 1" mean? If more commits are expected on this "merge data transfer" line, I would like to wait for them before starting testing within LMB.

arom4github:
I was informed that you prefer to have a minimal amount of modifications per pull request.  
That's why I planned to send yet three pull request similar to this. If you prefer, I can merge them into one and send them at once.

biochem-fan:
OK, probably I was not clear enough.

I don't like "squash merge" as I want to keep records of changes, but I prefer pull request to be grouped by logical blocks. In this case, please keep three commits as they are, but make only one pull request.

If you have done some tests (in terms of result or speed) on your side, please describe it as well.

arom4github:
Here they are.

The result is the same.

biochem-fan:
Thanks. We will test this next week. If we don't reply by the end of the next week, please remind us again.

biochem-fan:
I merged your contribution to our `nv-optim-datatransfer` branch. We will test this and then merge into the `ver3.1` branch.

biochem-fan:
Merged into `ver3.1`.

PR #581, opened 2020-02-26
Fix timers, optimize two kernels, fixed issue with pixel coordinate calculation

arom4github:
it's safer to use `blockDim.x` instead of a predefined constant.


biochem-fan:
Thanks. We will start internal testing.

biochem-fan:
We pushed this to LMB's cluster. We cannot put it GitHub yet because we also have several updates that need more internal testing. If you have more patches, please send the next one.

PR #588, opened 2020-03-12
PORT: Expectation setup  to  GPU

arom4github:
This part doesn't touch any FFT stuff. Moving FFT part to GPU will also reduce some data transfers.

### DIFF
--- a/src/acc/acc_helper_functions_impl.h
+++ b/src/acc/acc_helper_functions_impl.h
@@ -1869,6 +1869,22 @@ void windowFourierTransform2(
 	}
 }
 
+void run_calcPowerSpectrum(Complex *dFaux, int padoridim, Complex *ddata, int data_sz, RFLOAT *dpower_spectrum, RFLOAT *dcounter,
+											  int max_r2, int min_r2, RFLOAT normfft, RFLOAT padding_factor, RFLOAT weight,
+											  RFLOAT *dfourier_mask, int fx, int fy, int fz, bool do_fourier_mask)
+{
+#ifdef CUDA
+	dim3 bs(32,4,2);
+	dim3 gs(ceil((padoridim/2+1)/(float)bs.x), ceil(padoridim/(float)bs.y), ceil(padoridim/(float)bs.z));
+	if(sizeof(RFLOAT) == sizeof(double))
+		cuda_kernel_calcPowerSpectrum<<<gs,bs>>>((double2*)dFaux,padoridim,(double2*)ddata,data_sz,dpower_spectrum,dcounter,
+												  max_r2,min_r2,normfft,padding_factor,weight,dfourier_mask,fx,fy,fz,do_fourier_mask);
+	else
+		cuda_kernel_calcPowerSpectrum<<<gs,bs>>>((float2*)dFaux,padoridim,(float2*)ddata,data_sz,dpower_spectrum,dcounter,
+												  max_r2,min_r2,normfft,padding_factor,weight,dfourier_mask,fx,fy,fz,do_fourier_mask);
+	LAUNCH_HANDLE_ERROR(cudaGetLastError());
+#endif
+}
 
 
 void selfApplyBeamTilt2(MultidimArray<Complex > &Fimg, RFLOAT beamtilt_x, RFLOAT beamtilt_y,
--- a/src/acc/acc_helper_functions_impl.h
+++ b/src/acc/acc_helper_functions_impl.h
@@ -1886,6 +1886,14 @@ void run_calcPowerSpectrum(Complex *dFaux, int padoridim, Complex *ddata, int da
 #endif
 }
 
+void run_updatePowerSpectrum(RFLOAT *dcounter, int sz, RFLOAT *dpower_spectrum)
+{
+#ifdef CUDA
+	cuda_kernel_updatePowerSpectrum<<<ceil(sz/(float)256),256>>>(dcounter, dpower_spectrum, sz);
+	LAUNCH_HANDLE_ERROR(cudaGetLastError());
+#endif
+}
+
 
 void selfApplyBeamTilt2(MultidimArray<Complex > &Fimg, RFLOAT beamtilt_x, RFLOAT beamtilt_y,
 		RFLOAT wavelength, RFLOAT Cs, RFLOAT angpix, int ori_size)
--- a/src/acc/cuda/cuda_kernels/helper.cu
+++ b/src/acc/cuda/cuda_kernels/helper.cu
@@ -1002,3 +1002,14 @@ __global__ void cuda_kernel_griddingCorrect(RFLOAT *vol, int interpolator, RFLOA
 	}
 }
 
+__global__ void cuda_kernel_updatePowerSpectrum(RFLOAT *dcounter, RFLOAT *dpower_spectrum, int sz)
+{
+	int idx = blockIdx.x*blockDim.x + threadIdx.x;
+	if(idx<sz)
+	{
+		if (dcounter[idx] < 1.)
+			dpower_spectrum[idx] = 0.;
+		else
+			dpower_spectrum[idx] /= dcounter[idx];
+	}
+}
--- a/src/acc/cuda/cuda_kernels/helper.cu
+++ b/src/acc/cuda/cuda_kernels/helper.cu
@@ -491,70 +491,6 @@ __global__ void cuda_kernel_cosineFilter(	XFLOAT *vol,
 	}
 }
 
-__global__ void cuda_kernel_centerFFT_2D(XFLOAT *img_in,
-										 int image_size,
-										 int xdim,
-										 int ydim,
-										 int xshift,
-										 int yshift)
-{
-
-	int pixel = threadIdx.x + blockIdx.x*blockDim.x;
-	long int image_offset = image_size*blockIdx.y;
-//	int pixel_pass_num = ceilfracf(image_size, CFTT_BLOCK_SIZE);
-
-//	for (int pass = 0; pass < pixel_pass_num; pass++, pixel+=CFTT_BLOCK_SIZE)
-//	{
-		if(pixel<(image_size/2))
-		{
-			int y = floorf((XFLOAT)pixel/(XFLOAT)xdim);
-			int x = pixel % xdim;				// also = pixel - y*xdim, but this depends on y having been calculated, i.e. serial evaluation
-
-			int yp = (y + yshift + ydim)%ydim;
-			int xp = (x + xshift + xdim)%xdim;
-			int n_pixel = yp*xdim + xp;
-
-			XFLOAT buffer                  = img_in[image_offset + n_pixel];
-			img_in[image_offset + n_pixel] = img_in[image_offset + pixel];
-			img_in[image_offset + pixel]   = buffer;
-		}
-//	}
-}
-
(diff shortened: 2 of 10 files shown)


biochem-fan:
Thank you very much for your contribution.

I have two comments before merging this.

Even if the binary is compiled with CUDA, we cannot assume GPU is available at the run time (e.g. people run Class3D without alignment on a CPU node). We should not use GPU unless the user explicitly specified `--gpu`.

For centerFFT, we have just implemented a better way to do it. Instead of swapping elements in the real space, flip signs of every other elements in the reciprocal space (i.e. after FFT). Please see 0053ba1, 0697198 and 284137b. I think the same trick can be done on GPU.

arom4github:
it makes sense to rebase my changes on your updated branch

arom4github:
here it is

biochem-fan:
Thanks. Does this respect user's choice of the GPU device ID in `--gpu`?

arom4github:
Yes. It respects.

arom4github:
I have just checked the dataset you shared with us. It shows the same result for `ver3.1` branch and for this PR.
`FINAL RESOLUTION:              2.81827`

biochem-fan:
I will merge this into `devel-nvidia-optim3` for internal testing...

biochem-fan:
> src/acc/cuda/cuda_kernels/helper.cuh(985): error: no instance of overloaded func
tion "atomicAdd" matches the argument list
>            argument types are: (double *, double)

This does not compile. Do we always have `atomicAdd` for double?

arom4github:
the fix is there. I found you didn't use checks for '_ldg', so we skip this check as well.

The code was compiled with CUDA_ARCH=35

biochem-fan:
I am also using the default (CUDA_ARCH=35). The version of NVCC is 9.0, CPU double precision, GPU single precision.

arom4github:
Should I open a new PR?
I guess you can reopen this one to see the fix.

biochem-fan:
Unfortunately, it looks like a closed PR cannot be reopened.

But I will incorporate your commit https://github.com/arom4github/relion/commit/1eb0f25e4baa2e49f15557f64c98aa3f67adb257.

biochem-fan:
OK, compilation succeeded.

biochem-fan:
>> Does this respect user's choice of the GPU device ID in --gpu?
>
>Yes. It respects.

Is this because `cudaSetDevice` is called earlier in `ml_optimiser`?

When we make a GPU version of other programs (say `relion_project`), we should call `cudaSetDevice` before using `Projector`. Is this correct?
(This is completely fine; I just want to understand the assumption)

arom4github:
Yes. I inserted `cudaSetDevice` here (see `src/ml_optimiser_mpi.cpp`) 
https://github.com/3dem/relion/pull/588/files#diff-1e0d3c2ed6ceb8c74bc84b4698a33a35R298

biochem-fan:
Thanks. I deployed this to our internal cluster. If nobody find issues for one week, I will merge it into `ver3.1`.

biochem-fan:
Unfortunately, your change broke the build for CPU nodes...

When compiled with CUDA, the binary no longer runs on a node without GPU.
When I set `CUDA_VISIBLE_DEVICES=""`, the program dies as follows, even without `--gpu`.

I guess the use of `CudaCustomAllocator` is the problem.

```
=== Backtrace  ===
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN11RelionErrorC1ERKSsS1_l+0x41) [0x442261]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi() [0x5b99c9]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN9Projector26computeFourierTransformMapER13MultidimArrayIdES2_iibbiPKS1_b+0x22b8) [0x5bd538]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN7MlModel23setFourierTransformMapsEbidPK13MultidimArrayIdE+0x9d2) [0x4f3a22]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN11MlOptimiser16expectationSetupEv+0x5e) [0x5eb27e]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN14MlOptimiserMpi11expectationEv+0x551) [0x468931]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(_ZN14MlOptimiserMpi7iterateEv+0xc1) [0x4783f1]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi(main+0x5f) [0x43248f]
/lib64/libc.so.6(__libc_start_main+0xf5) [0x7fa2f0e2d495]
/lmb/home/tnakane/prog/relion-devel-lmb/build-test/bin/relion_refine_mpi() [0x4361af]
==================
ERROR: 

A GPU-function failed to execute.
```

arom4github:
Ops!
That improves my way of testing.
The fix is here arom4github:optim3

biochem-fan:
I applied your fix and it seems OK now. I re-deployed it to our cluster.

biochem-fan:
Did you try this on Class2D? It crashes.

Note that for Class2D the reference is 2D, not 3D. So the size of Faux is smaller.

```
#0  0x00007fffe5b88158 in ?? () from /lib64/libcuda.so.1
#1  0x00007fffe5c4393e in ?? () from /lib64/libcuda.so.1
#2  0x00007fffe5d30b17 in ?? () from /lib64/libcuda.so.1
#3  0x00007fffe5c4504e in ?? () from /lib64/libcuda.so.1
#4  0x00007fffe5b530ce in ?? () from /lib64/libcuda.so.1
#5  0x00007fffe5b55243 in ?? () from /lib64/libcuda.so.1
#6  0x00007fffe5cbb562 in cuMemcpyHtoDAsync_v2 () from /lib64/libcuda.so.1
#7  0x00007fffed54f8cc in ?? () from /public/EM/CUDA/cuda-8.0/lib64/libcudart.so.8.0
#8  0x00007fffed52bb5b in ?? () from /public/EM/CUDA/cuda-8.0/lib64/libcudart.so.8.0
#9  0x00007fffed565b08 in cudaMemcpyAsync () from /public/EM/CUDA/cuda-8.0/lib64/libcudart.so.8.0
#10 0x00007ffff75da1eb in CudaShortcuts::cpyHostToDevice<tComplex<float> > (h_ptr=0x1d0a960, d_ptr=0x7fffac000000, 
    size=8454144, stream=0x0) at src/acc/cuda/shortcuts.cuh:32
#11 0x00007ffff75db9c4 in AccPtr<tComplex<float> >::cpToDevice (this=0x7fffffffaa60)
    at src/acc/acc_ptr.h:523
#12 0x00007ffff75d3f88 in Projector::computeFourierTransformMap (this=0xaed7a0, vol_in=..., power_spectrum=..., 
    current_size=38, nr_threads=6, do_gridding=true, do_heavy=true, min_ires=-1, fourier_mask=0x0, do_gpu=true)
    at src/projector.cpp:245
#13 0x00007ffff78970b0 in MlModel::setFourierTransformMaps (this=0x7fffffffbb30, update_tau2_spectra=true, 
    nr_threads=6, strict_lowres_exp=-1, fourier_mask=0x0)
    at src/ml_model.cpp:1375
```

arom4github:
I fixed this. you can take it here `arom4github:optim3`.
I can test it myself first, but I hope you have smaller dataset for testing it. At least it passed 2 iterations.

arom4github:
Hi,
I have pushed the fix.
now it gives me the correct result for 2D and 3D reconstruction

biochem-fan:
Thanks. We will release 3.1.0 first and then incorporate this in 3.1.1.

biochem-fan:
Class2D worked at least on my test case. Deployed to our cluster.

biochem-fan:
@arom4github Unfortunately, this crashes with `--pad 1`.

```KERNEL_ERROR: invalid argument in /public/EM/RELION/relion-3.1-dev/src/acc/utilities_impl.h at line 719 (error-code 11)```

arom4github:
It's a silly error. Never falt into this case.  Hope, I have all cases to test before submitting the next PR.
The fix was pushed.

biochem-fan:
Just to make sure: is this safe with a very large box (768 or even 1024 px) with 11 GB of VRAM? This is the case where pad 1 is used.

arom4github:
How can I test it?  
I put this option for 3D reconstruction (Plasma ribosome dataset), and it passed 4 iterations well (8xV100, 4 MPI ranks per GPU)

biochem-fan:
Isn't this function sending doubles to GPU? In contrast to `XFLOAT`, `RFLOAT` is double by default.

If we have a 1024 px box at `--pad 1` (or 512 px box at `--pad 2`), it is 1024 * 1024 * 1024 * 8 = 8 GB. If we have something else on GPU, it will probably run out of memory.

arom4github:
According to the RELION workflow, we don't have anything else on GPU at that time. The only case to have something there it to have another application which uses GPU. But in this case, you can run into the same problem in the Expectation part.

BTW, what are the minimal requirements for GPU? Compute capability, GPU mem?

biochem-fan:
> BTW, what are the minimal requirements for GPU? Compute capability, GPU mem?

This depends on the dataset. For a larger box, we need larger memory. Currently, 1080 Ti with 11 GB RAM can process  up to a 1,000 px box with `--pad 1`, which is large enough for most cases.

I guess most users use GPUs with CC 5.2 or later but we need to discuss if we are to drop older cards.

bforsbe:
All accelerated code introduced up until 2.0 utilizes functionalities introduced  with cc 3.5, so that is the formal requirement. Not sure if other developement since use more recent feature sets. 

Also, as a comment on earlier mentions, we always use XFLOAT in GPU-kernels, and cast the input/output. I think there's even a preparatory kernel that perhaps somewhat stupidly casts an entire array...

biochem-fan:
I deployed this to our cluster.

arom4github:
For the projection part (without FFT for now), it's required  1 GB with `--pad 1`, 5.9GB with `--pad 2` (or without this option).
I'm confused here, I thought you expected to have bigger size with smaller `--pad` option.

Did I get it right that it's possible carrying out the same calculations in `float` (Expectation setup, maximization)? Do you expect to have the same result in terms of accuracy?

biochem-fan:
> I thought you expected to have bigger size with smaller `--pad` option.

No. If the input is 100 x 100 x 100 px, `--pad 1` performs FFT on it. `--pad 2` means padding the input to 200 x 200 x 200 px in the real space and then performs FFT. Thus, `--pad 2` takes 8 times more memory.

> Did I get it right that it's possible carrying out the same calculations in float (Expectation setup, maximization)? Do you expect to have the same result in terms of accuracy?

I guess so, but we don't want to change numerical precision at this moment. One place double is absolutely necessary is gridding, but it is no longer default (`--skip_gridding` is on in RELION 3.1).

arom4github:
I guess you may forget the last commit https://github.com/arom4github/relion/commit/497453af8ec78f77facd16d8606955f201c3c8ce

biochem-fan:
This looks OK now. I merged this into `ver3.1`. This will be included in 3.1.1.

Issue #654, opened 2020-08-09
GPU Mem Use clarification


craigyk:
This is probably not a bug, but my last recollection was that it should be possible to use a box size of ~1000x1000 in 11GB of GPU mem with a padding of 1.  Is this no longer the case (or was it ever?).  I added some printf statements once I saw CUDAMemAlloc errors, and noticed with a box size of 1024, Relion was trying to allocated 16GB of GPU RAM.  I would have guessed each map FFT in float would have been ~8GB, but it seems to be trying to grab twice that.  Is that for each half map?  I thought the independent halfsets were split between different MPI processes?

Thanks for clarifications!

biochem-fan:
1000 px should be OK with `--pad 1`. Are you sure you built with GPU single precision (the default)?

Are you using version 3.1.0 (1e738ec) or something newer in the ver3.1 branch?

craigyk:
Yes, I'm using the latest commits to the v3.1 branch

craigyk:
I am reverting to 3.1.0 tag and trying that...

bforsbe:
There's a weight array for the weighted back-projection, so you have to bump the estimate by 50%, putting you at 12GB, unless you accounted for that.

craigyk:
> There's a weight array for the weighted back-projection, so you have to bump the estimate by 50%, putting you at 12GB, unless you accounted for that.

I didn't account for that, but I also forgot that the FFT of a real-valued volume would not need 2X the memory.  In any case, it seems to be running great again with 11GB after reverting to 3.1.0

biochem-fan:
@arom4github Can you investigate this issue? Probably your new Projector/Backprojector initialization code is the cause. I would revert your commit if this cannot be fixed.

arom4github:
Hi,

it's easy to fix. 
In a day or so I will push the next PR. It will bring FFT stuff for the projection (Expectation setup) on GPU and fix this issue.
I already have a code which checks available GPU memory and in case there is not enough mem on the GPU, it performs calculations on CPU side.
 
I didn't have a dataset to test such big images. Where can I get it?

Alexey

biochem-fan:
Thanks.

> I didn't have a dataset to test such big images. Where can I get it?

1. Make the reference large:
  ```relion_image_handler --i ref.mrc --o ref_1K.mrc --new_box 1000```
2. Use it in`relion_refine-mpi` with `--ref ref_1K.mrc --pad 1 --trust_ref_size`.

[...]

PR #657, opened 2020-08-13
PORT: bring FFT stuff for the projection to GPU

arom4github:
This PR:
* performs FFT on GPU for the Expectation setup part
* fixes the issue discussed here https://github.com/3dem/relion/issues/654#issuecomment-671171462
* fixes several related CUDA kernels in terms of not taking predefined constant for the block size.

### DIFF
--- a/src/projector.cpp
+++ b/src/projector.cpp
@@ -20,12 +20,14 @@
 #include "src/projector.h"
 #include "src/jaz/gravis/t3Vector.h"
 #include <src/time.h>
+#ifdef CUDA
 #include <cufft.h>
 #include <sys/types.h>
 #include <sys/ipc.h>
 #include <sys/sem.h>
 #define SEMKEY 826976737978L /* key value for semget() */
 #define PERMS 0666
+#endif
 //#define DEBUG
 
 //#define PROJ_TIMING
--- a/src/projector.cpp
+++ b/src/projector.cpp
@@ -206,7 +206,6 @@ void Projector::computeFourierTransformMap(
 			allocator = new CudaCustomAllocator(mem_req, (size_t)16);
 		else
 		{
-			std::cerr << " WARNING: There is not enough memory to process the Expectation setup stage on GPU. It will be executed on CPU." << std::endl;
 			do_gpu = false; // change local copy of do_gpu variable
 			if (semop(semid, &op_unlock[0], 1) < 0)
 				REPORT_ERROR("semop unlock error");
--- a/src/projector.cpp
+++ b/src/projector.cpp
@@ -20,6 +20,12 @@
 #include "src/projector.h"
 #include "src/jaz/gravis/t3Vector.h"
 #include <src/time.h>
+#include <cufft.h>
+#include <sys/types.h>
+#include <sys/ipc.h>
+#include <sys/sem.h>
+#define SEMKEY 826976737978L /* key value for semget() */
+#define PERMS 0666
 //#define DEBUG
 
 //#define PROJ_TIMING
@@ -152,22 +158,58 @@ void Projector::computeFourierTransformMap(
 		REPORT_ERROR("Projector::computeFourierTransformMap%%ERROR: Dimension of the data array should be 2 or 3");
 	}
 #ifdef CUDA
-	size_t mem_req;
+	static int semid = -1;
+	struct sembuf op_lock[2]=  { 0, 0, 0, /* wait for sem #0 to become 0 */
+	     						 0, 1, SEM_UNDO /* then increment sem #0 by 1 */ };
+	struct sembuf op_unlock[1]= { 0, -1, (IPC_NOWAIT | SEM_UNDO) /* decrement sem #0 by 1 (sets it to 0) */ };
+
+	size_t mem_req, ws_sz;
 	int Faux_sz = padoridim*(padoridim/2+1);
+	int n[3] = {padoridim, padoridim, padoridim};
+	cufftType cufft_type = CUFFT_R2C;
+	if(sizeof(RFLOAT) == sizeof(double))
+		cufft_type = CUFFT_D2Z;
 	
 	if(ref_dim == 3)
 		Faux_sz *= padoridim;
 
 	mem_req =  (size_t)1024;
 	if(do_heavy && do_gpu)
+	{
+	    cufftEstimateMany(ref_dim, n, NULL, 0, 0, NULL, 0, 0, cufft_type, 1, &ws_sz);
+
 		mem_req = (size_t)sizeof(RFLOAT)*MULTIDIM_SIZE(vol_in) +                   // dvol
 				  (size_t)sizeof(Complex)*Faux_sz +                                // dFaux
-				  (size_t)sizeof(RFLOAT)*MULTIDIM_SIZE(Mpad);                      // dMpad
+				  (size_t)sizeof(RFLOAT)*MULTIDIM_SIZE(Mpad) +                     // dMpad
+				  ws_sz + 16;                                                      // worspace for cuFFT
+	}
 
 	CudaCustomAllocator *allocator = NULL;
-	if(do_gpu) 
-		allocator = new CudaCustomAllocator(mem_req, (size_t)16);
+	if(do_gpu && do_heavy)
+	{
+		int devid;
+		size_t mem_free, mem_tot;
+		cudaDeviceProp devProp;
+		if (semid <0) 
+		{
+			HANDLE_ERROR(cudaGetDevice(&devid));
+			if ( ( semid=semget(SEMKEY+devid, 1, IPC_CREAT | PERMS )) < 0 ) 
+				REPORT_ERROR("semget error"); 
+		}
+		if (semop(semid, &op_lock[0], 2) < 0) 
+			REPORT_ERROR("semop lock error");
 
+		HANDLE_ERROR(cudaMemGetInfo(&mem_free, &mem_tot));
+		if(mem_free > mem_req)
+			allocator = new CudaCustomAllocator(mem_req, (size_t)16);
+		else
+		{
+			std::cerr << " WARNING: There is not enough memory to process the Expectation setup stage on GPU. It will be executed on CPU." << std::endl;
+			do_gpu = false; // change local copy of do_gpu variable
+			if (semop(semid, &op_unlock[0], 1) < 0)
+				REPORT_ERROR("semop unlock error");
+		}
+	}
 	AccPtrFactory ptrFactory(allocator);
 	AccPtr<RFLOAT> dMpad = ptrFactory.make<RFLOAT>(MULTIDIM_SIZE(Mpad));
 	AccPtr<Complex> dFaux = ptrFactory.make<Complex>(Faux_sz);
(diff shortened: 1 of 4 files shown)


src/projector.cpp
```diff
@@ -20,6 +20,12 @@
 #include "src/projector.h"
 #include "src/jaz/gravis/t3Vector.h"
 #include <src/time.h>
+#include <cufft.h>
```
biochem-fan, line 23: Please keep the ability to compile RELION without CUDA SDK by guarding by `#ifdef CUDA`.

src/projector.cpp
```diff
@@ -152,22 +158,58 @@ void Projector::computeFourierTransformMap(
...
-		allocator = new CudaCustomAllocator(mem_req, (size_t)16);
+	if(do_gpu && do_heavy)
+	{
+		int devid;
+		size_t mem_free, mem_tot;
+		cudaDeviceProp devProp;
+		if (semid <0) 
+		{
+			HANDLE_ERROR(cudaGetDevice(&devid));
+			if ( ( semid=semget(SEMKEY+devid, 1, IPC_CREAT | PERMS )) < 0 )
```
biochem-fan, line 196: In RELION, we use `pthread_mutex_lock` (see `ml_optimiser.cpp`) or `#pragma omp critical(CRITICAL_SECTION_NAME` for this purpose.

arom4github: hi,
that's not the case. I need to sync MPI ranks which assigned to the same GPU. I can't use OMP critical section or mutex as I need to sync processes, not threads.
Once you prefer not to use semaphores, I can rewrite the code  to use MPI calls

src/projector.cpp
arom4github: fixed

src/projector.cpp
biochem-fan: I see. OK, then go ahead with semaphores. You shouldn't use MPI call here because this class is also used from non-MPI programs.

biochem-fan:
Some confirmations before merging:

- Did you actually try compilation without CUDA?
- Did you make sure this works with `--pad 1` and `--pad 2`?
- How about CPU single precision build? `cmake -DDoublePrec_CPU=OFF`
- Did you try 1000 px box? Did my command in #654 work? Otherwise I will tell you have to up-sample particles.

arom4github:
* yes. the code compiled smoothly without CUDA.
* yes. I tried `--pad 1` and `--pad 2`. 
* yes. the code compiled well with `-DDoublePrec_CPU=OFF`. For the small dataset you shared I see ` FINAL RESOLUTION:              2.81827`
* I was able to create ref_1k file, but even `ver3.1` branch dies having it as an input. To test memory limitations I requested twice more memory for the ordinary dataset. Result - switching back to CPU calculation for the Expectation setup.

biochem-fan:
> I was able to create ref_1k file, but even ver3.1 branch dies having it as an input. 

Please try tag `3.1.0` (i.e. the latest in `master`). It works.
```
relion_image_handler --i ref.mrc --o ref_1K.mrc --new_box 1000
relion_image_handler --i mask.mrc --o mask_1K.mrc --new_box 1000
mpirun -np 5 relion_refine_mpi --o Refine3D/job001/run1Kpx --auto_refine --split_random_halves --i Polish/job028/shiny.star --ref ref_1K.mrc --firstiter_cc --ini_high 40 --dont_combine_weights_via_disc --scratch_dir /ssd --pool 3 --pad 1  --skip_gridding  --auto_ignore_angles --auto_resol_angles --ctf --ctf_corrected_ref --particle_diameter 200 --flatten_solvent --zero_mask --solvent_mask mask_1K.mrc --solvent_correct_fsc  --oversampling 1 --healpix_order 2 --auto_local_healpix_order 4 --offset_range 5 --offset_step 2 --sym D2 --low_resol_join_halves 40 --norm --scale  --j 8 --gpu "" --trust_ref_size
```

This is before merging your `optim3` PR, while the latest in `ver3.1` branch already contains it.

biochem-fan:
Actually, with this `optim4` patch, the 1K test above works! I will ask @craigyk to test on his real dataset.

src/projector.cpp
```diff
@@ -152,22 +160,58 @@ void Projector::computeFourierTransformMap(
...
+		}
+		if (semop(semid, &op_lock[0], 2) < 0) 
+			REPORT_ERROR("semop lock error");
 
+		HANDLE_ERROR(cudaMemGetInfo(&mem_free, &mem_tot));
+		if(mem_free > mem_req)
+			allocator = new CudaCustomAllocator(mem_req, (size_t)16);
+		else
+		{
+			std::cerr << " WARNING: There is not enough memory to process the Expectation setup stage on GPU. It will be executed on CPU." << std::endl;
```
biochem-fan, line 209: Please remove this warning, because this is harmless (i.e. just use the old code path).

biochem-fan:
Unfortunately, this still causes out-of-memory errors.

One of our users reported the following (reproducible) crash in Refine3D. 1 MPI process / 2080Ti. The box size is 330 px, so should fit in the GPU memory.

```
[fmg11:1489057] *** Process received signal ***
[fmg11:1489057] Signal: Segmentation fault (11)
[fmg11:1489057] Signal code: Address not mapped (1)
[fmg11:1489057] Failing at address: 0x28
[fmg11:1489057] [ 0] /lib64/libpthread.so.0(+0xf5d0)[0x2b1d52d4b5d0]
[fmg11:1489057] [ 1] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN6AccPtrIcE9freeIfSetEv+0x46)[0x47c996]
[fmg11:1489057] [ 2] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN9Projector26computeFourierTransformMapER13MultidimArrayIdES2_iibbiPKS1_b+0x3784)[0x478584]
[fmg11:1489057] [ 3] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN7MlModel23setFourierTransformMapsEbidPK13MultidimArrayIdE+0x902)[0x625f42]
[fmg11:1489057] [ 4] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN11MlOptimiser16expectationSetupEv+0x5e)[0x54273e]
[fmg11:1489057] [ 5] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN14MlOptimiserMpi11expectationEv+0x551)[0x446b51]
[fmg11:1489057] [ 6] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(_ZN14MlOptimiserMpi7iterateEv+0xc1)[0x456611]
[fmg11:1489057] [ 7] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi(main+0x5f)[0x4328ef]
[fmg11:1489057] [ 8] /lib64/libc.so.6(__libc_start_main+0xf5)[0x2b1d52f7a3d5]
[fmg11:1489057] [ 9] /public/EM/RELION/relion-3.1-dev/build/bin/relion_refine_mpi[0x43664f]
[fmg11:1489057] *** End of error message ***
in: /public/EM/RELION/relion-3.1-dev/src/acc/cuda/custom_allocator.cuh, line 546
ERROR: 

You ran out of memory on the GPU(s).

Each MPI-rank running on a GPU increases the use of GPU-memory. Relion
tries to distribute load over multiple GPUs to increase performance,
but doing this in a general and memory-efficient way is difficult.

1. Check the device-mapping presented at the beginning of each run,
   and be particularly wary of 'device X is split between N slaves', which 
   will result in a higher memory cost on GPU X. In classifications, GPU-
   sharing between MPI-ranks is typically fine, whereas it will usually 
   cause out-of-memory during the last iteration of high-resolution refinement.

2. If you are not GPU-sharing across MPI-slave ranks, then you might be using a
   too-big box-size for the GPU memory. Currently, N-pixel particle images
   will require *roughly* 

		    (1.1e-8)*(N*2)^3  GB  

   of memory (per rank) during the final iteration of refinement (using
   single-precision GPU code, which is default). 450-pixel images can therefore
   just about fit into a GPU with 8GB of memory, since 11*(450*2)^3 ~= 8.02
   During classifications, resolution is typically lower and N is suitably
   reduced, which means that memory use is much lower.

3. If the above estimation fits onto (all of) your GPU(s), you may have 
   a very large number of orientations which are found as possible during
   the expectation step, which results in large arrays being needed on the 
   GPU. If this is the case, you should find large (>10'000) values of 
   '_rlnNrOfSignificantSamples' in your _data.star output files. You can try
   adding the --maxsig <P>, flag, where P is an integer limit, but you 
   should probably also consult expertise or re-evaluate your data and/or 
   input reference. Seeing large such values means relion is finding nothing
   to align.

If none of the above applies, please report the error to the relion
developers at    github.com/3dem/relion/issues
```

biochem-fan:
It also says:

```
ERROR: CudaCustomAllocator out of memory
 [requestedSpace:             2306937600 B]
 [largestContinuousFreeSpace: 2306937600 B]
 [totalFreeSpace:             2594433600 B]
[287496000B] (2299968000B) (2306937600B) [2306937600B] = 7201339200B
```

arom4github:
Can you reproduce this issue? I failed to do this. I tried to create ref and mask files for the box of size 330 as you pointed above for 1K images.

According to the size of the requested memory, it is `dFaux` variable. Before the allocation of this variable, there should be enough space to allocate it. Moreover, there should be free space for FFT plan also.

craigyk:
I did not have any further memory problems with this revision- but I also realized I had turned off the cached allocator in the build.

biochem-fan:
@arom4github At least one user could reproduce it. After crashing, she submitted exactly the same job, which crashed again. Because this involves unpublished data, I cannot share it to you.

Does `cudaMemGetInfo` guarantee that you can allocate the amount as a continuous stretch of memory even with a CustomAllocator?

arom4github:
> Does cudaMemGetInfo guarantee that you can allocate the amount as a continuous stretch of memory even with a CustomAllocator?
yes.  At least according to the output, the CustomAllocator allocates the memory but failed to allocate space for the variable inside the allocated chunk.

arom4github:
It's difficult to say what is the reason for the memory allocation error having no way to access the dataset. Anyway, I made some assumptions and extended the size of the requested memory. Can you, please, test it? Definitely, there are a few other ways to fix this.

Meanwhile,  https://github.com/3dem/relion/blob/master/src/acc/cuda/custom_allocator.cuh#L125
Why do you use `a->size <= size` here? is it to have at least a one-byte gap between variables?

biochem-fan:
> the CustomAllocator allocates the memory but failed to allocate space for the variable inside the allocated chunk.

Does this mean a bug within CustomAllocator?

> Why do you use a->size <= size here? is it to have at least a one-byte gap between variables?

I don't know. This was written by @dkimanius or @bforsbe.

arom4github:
> Does this mean a bug within CustomAllocator?

nope. It's could be the feature of CustomAllocator. it ignores the chanks which exactly match the requested size.

biochem-fan:
Your latest patch (`+ 4096`) fixed the crash. I will deploy this to our cluster and see what happens.

biochem-fan:
This looks OK now. I merged it into the ver3.1 branch to be released as 3.1.1.

arom4github:
That's cool! thanks!

PR #670, opened 2020-09-06
Fix typo in helpstring

smsaladi:
I think the helpstring should read as follows, specifying how to turn off early_binning ...


biochem-fan:
You are right. Thanks for pointing this.

I will merge this into ver3.1 branch, which will soon be version 3.1.1.

PR #671, opened 2020-09-09
Devel altcpu 2dopt

CharlesCongdon:
Merge latest  altcpu optimizations into 3.1


CharlesCongdon:
Results look reasonable in 2D and 3D for Plasmodium.  Could you check to see if everything looks OK on your end and add to 3.1 if OK?

PR #678, opened 2020-09-18
PORT: bring FFT stuff to GPU for backprojection

arom4github:
this PR brings cuFFT for the `back-projection` stage. We also optimized data locality and removed some redundant sync points.


martin-g:
The PR could be closed and then reopened to trigger the CI checks.

biochem-fan:
Actually we can close this PR because we no longer use gridding (private communication).

Issue #705, opened 2020-11-03
Image index prefix in filename with mrc extension


joton:
When input is a starfile, relion_image_handler composes the output filename by adding the index number prefix even if output image file is containing one single image instead of a stack. There's no bug here. However, when that output starfile is used as input with images being 3d subtomograms, Image class throws an error trying to read them because an image index is passed and extension is not MRCS.

Which strategy should we follow to fix it? to correct the output starfile in image_handler and other programs as they show up or allowing Image class to read files with select_img=0 (instead of -1) and mrc extension? I think the former should be fixed to be coherent with input. The latter would be make image read more flexible but I don't know if that would break something elsewhere.

biochem-fan:
Can you give me examples (input file names, output file names, etc) of the problems  and proposed two solutions?

joton:
Sure!
 
The input star file contains file names like "subtomograms/ini_10k/0_data.mrc"

After running the command "relion_image_handler --i subtomograms/ini_10k_particles_bug.star --o DELETEME --phase_randomise 6" I get the files properly processed and the output star file "subtomograms/ini_10k_particles_bug_DELETEME.star" whose image file names are now like "000001@subtomograms/ini_10k/0_data_DELETEME.mrc".

To fix relion_image_handler we'd need to modify a couple of blocks: we should check whether real image files within the starfile are stacks instead of setting input_is_stack to True directly (image_handler.cpp:718), replacing by

FileName fn_img;
MD.getValue(EMDL_IMAGE_NAME, fn_img, 0);
fn_img.decompose(slice_id, fn_stem);
input_is_stack = (fn_in.getExtension() == "mrcs" || fn_in.getExtension() == "tif" || fn_in.getExtension() == "tiff") && (slice_id == -1);

Then, for my_fn_out composition in lines 980-991 we must consider when input_is_star is True while input_is_stack is False inserting the suffix without adding the index number prefix.

Fixing image class is far simpler. We could read .mrc extension filenames with index number prefix if index == 1, replacing in image.h:1397 " select_img >= 0" by "select_img >0"

Line references are for master branch at current commit (89480b1). I tested both changes and they fix my problem. I could commit them but, although the logic looks quite simple, I prefer to follow your suggestions.

biochem-fan:
Thanks for your explanation. Now it is clear.

[...]

PR #709, opened 2020-11-10
Fixes image numbering in filename for files other than stacks in starfiles #705

joton:
Hi @biochem-fan ,

I've did several tests with a particles.star file from the tutorial. I tried all the possible combinations of outputs and it works as expected.


biochem-fan:
This looks OK. I merged it. Thank you very much.

joton:
Thanks to you too!!

Issue #731, opened 2021-02-09
Builds on Ubuntu Bionic (gcc@7.5.0) Fail Due to Link Order


bollig:
Building relion@3.1.1 on ubuntu18.04 with gcc@7.5.0. Enabled options: MKLFFT=on, DoubleCPU=ON, CUDA=Off. 

Build fails with: 

```
[ 61%] Linking CXX executable ../../bin/relion_star_handler
cd /tmp/ubuntu/spack-stage/spack-stage-relion-3.1.1-walxkdxa77vholpzwnmc4cfodqcyi5ls/spack-build-walxkdx/src/apps && /shared/spack-0.16/opt/spack/linux-ubuntu18.04-skylake_avx512/gcc-7.5.0/cmake-3.13.5-kfs37giuvmqvbfjo7h435jw7sr7pruqd/bin/cmake -E cmake_link_script CMakeFiles/star_handler.dir/link.txt --verbose=1
[...]

PR #732, opened 2021-02-09
Simple patch to fix intel mkl linking

bollig:
Quick hack, but it ensures that MKL is linked properly when MKLFFT is on but compiler is GCC. A better approach would be to add FindMKL.cmake (see e.g., https://github.com/pytorch/pytorch/blob/master/cmake/Modules/FindMKL.cmake), and always include your MKL libraries via target_link_libraries().


bollig:
Fixes #731

biochem-fan:
Repaired by @do-jason's 58bffcb as discussed in #731.

PR #739, opened 2021-02-23
Updated code with inclusive terms leader and follower

mcianfrocco:
Hello RELION developers - 

Related to discussions of how to make science and cryoEM inclusive, we should consider removing 'master' and 'slave' from our code. To help facilitate this, I'm submitting this pull request where I replaced 'master' and 'slave' with 'leader' and 'follower,' respectively.

Some additional materials discussing this situation in the life sciences & computing:

- [Equity, Diversity and Inclusion: A call to eradicate non-inclusive terms from the life sciences](https://elifesciences.org/articles/65604)
- [Tech Confronts Its Use of the Labels ‘Master’ and ‘Slave’](https://www.wired.com/story/tech-confronts-use-labels-master-slave/)

I've tested the changes on my end without any issues in analyzing the tutorial dataset. 

If you prefer alternative terms I can also change and submit a different pull request. 

Thank you!
Mike

Screen Shot 2021-02-23 at 12 27 50 PM

### DIFF
--- a/src/autopicker_mpi.cpp
+++ b/src/autopicker_mpi.cpp
@@ -25,17 +25,17 @@ void AutoPickerMpi::read(int argc, char **argv)
 	// Define a new MpiNode
 	node = new MpiNode(argc, argv);
 
-	if (node->isMaster())
+	if (node->isLeader())
 		PRINT_VERSION_INFO();
 
 	// First read in non-parallelisation-dependent variables
 	AutoPicker::read(argc, argv);
 
-	// Don't put any output to screen for mpi slaves
-	if (!node->isMaster())
+	// Don't put any output to screen for mpi followers
+	if (!node->isLeader())
 		verb = 0;
 
-	if (do_write_fom_maps && node->isMaster())
+	if (do_write_fom_maps && node->isLeader())
 		std::cerr << "WARNING : --write_fom_maps is very heavy on disc I/O and is not advised in parallel execution. If possible, using --shrink 0 and lowpass makes I/O less significant." << std::endl;
 
 	// Possibly also read parallelisation-dependent variables here
@@ -60,9 +60,9 @@ int AutoPickerMpi::deviceInitialise()
 	else
 		dev_id = textToInteger((allThreadIDs[node->rank][0]).c_str());
 
-	for (int slave = 0; slave < node->size; slave++)
+	for (int follower = 0; follower < node->size; follower++)
 	{
-		if (slave == node->rank)
+		if (follower == node->rank)
 		{
 			std::cout << " + Using GPU device: " << dev_id << " on MPI node: " << node->rank << std::endl;
 			std::cout.flush();
--- a/src/ctffind_runner_mpi.cpp
+++ b/src/ctffind_runner_mpi.cpp
@@ -27,8 +27,8 @@ void CtffindRunnerMpi::read(int argc, char **argv)
 	// First read in non-parallelisation-dependent variables
 	CtffindRunner::read(argc, argv);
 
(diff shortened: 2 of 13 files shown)


biochem-fan:
Since I am not a native speaker of English, I cannot properly evaluate the connotation of these terms. I will leave decision to @scheres.

eriklindahl:
For GROMACS we're trying to move to server/client, but the exact nomenclature isn't that critical, I think.

One can of course easily argue a particular change in a particular source code won't have huge impact in the world, and that people have many other priorities, but most of us are also privileged and have never felt the injustice of history. In particular when @mcianfrocco has already done the work I would strongly suggest we simply merge this to do our part.

biochem-fan:
I rebased the patch to `ver3.1`, which is the latest (public) development branch and will be 3.1.2 in near future.

mcianfrocco:
Thank you @biochem-fan & @eriklindahl !

scheres:
Thanks for bringing this up Mike. I don't really like leader/follower. Would replacing 'slave' by 'worker', and leaving master as it is be an alternative?

eriklindahl:
We've discussed that in other occasions (codes), and the argument there has been that when the two have occurred as a pair for 30+ years, it looks half-hearted to only replace one of them (since the word definitely has been referring to a certain type of "master"). I would suggest either server-client, server-worker,  director-worker or controller-worker to just avoid that.

biochem-fan:
@mcianfrocco

Sorry, I was not clear. I meant I changed the target of the patch to `ver3.1` by "rebased"; in other words, the patch is not yet merged. The discussion is ongoing. I reopened this PR.

delarosatrevin:
What about manager/worker?

scheres:
@mcianfrocco did you need to do anything else than a sed operation for master, Master, slave & Slave on the entire code base? Any pitfalls there?

scheres:
Hmmm, I quite like worker, but don't really like the alternatives for rank=0. Thinking about what is going on a bit more, perhaps leader/follower is actually not so bad: it is more symmetric and less emotional. Another advantage is that it keeps @mcianfrocco 's suggestion, who brought it up in the first place.  Will merge now. Thanks again Mike.

biochem-fan:
> I've tested the changes on my end without any issues in analyzing the tutorial dataset.

This patch was incomplete.

```diff
diff --git a/src/acc/cuda/cuda_autopicker.cu b/src/acc/cuda/cuda_autopicker.cu
index e52309a..61467b4 100644
--- a/src/acc/cuda/cuda_autopicker.cu
+++ b/src/acc/cuda/cuda_autopicker.cu
@@ -77,7 +77,7 @@ AutoPickerCuda::AutoPickerCuda(AutoPickerMpi *basePicker, int dev_id, const char
 
 {
        node = basePicker->getNode();
-       basePicker->verb = (node->isMaster()) ? 1 : 0;
+       basePicker->verb = (node->isLeader()) ? 1 : 0;
 
        projectors.resize(basePckr->Mrefs.size());
        have_warned_batching=false;
```

biochem-fan:
Actually there were **many** remaining occurrences in the source codes. I replaced them but please make a complete pull request next time.

mcianfrocco:
Thank you @biochem-fan - I apologize for missing these instances. 

One version that came from Ben Himes on Twitter was Dispatcher / Agent. 

Thank you for considering & including these changes! 
mike

asntech:
Thanks @mcianfrocco @biochem-fan @scheres and RELION developers/contributors for taking such quick action/decision to replace the problematic terms!! 👏🏼

PR #756, opened 2021-03-25
Don't use MPI wrapper as compiler if using Intel compilers.

tjragan:
6918e6d945b618d86314b8dbb339cc240b3c8861 breaks compilation with Intel compilers.  This seems to fix it.


CharlesCongdon:
T.J.:
    Do you find such changes are necessary if you build under ICC with "cmake -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc -DMPI_C_COMPILER=mpiicc -DMPI_CXX_COMPILER=mpiicpc ..." rather than "CC=mpiicc CXX=mpiicpc cmake ...."?
    The build documentation is a little behind the latest code, although https://relion.readthedocs.io/en/latest/Installation.html is now updated.
    Thanks,
    Charles

do-jason:
@tjragan Did you tried cmake configuration suggested by @CharlesCongdon ?
I've tested some cmake configurations and I guess you have used mpicc/mpiicpc as C/C++ compiler command in cmake. And if this is the case, then I guess your provided fix also does not work properly in cmake configuration.
If not, please let me know your cmake configuration command and error messages.

To make cmake more solid for Intel compiler, the below change will provide workaround for users who specify mpiicc/mpiicpc as C/C++ compiler.

```cmake
if ("${MPI_CXX_INCLUDE_DIRS}" STREQUAL "")
        include_directories("${MPI_CXX_INCLUDE_PATH}")
else()
        include_directories("${MPI_CXX_INCLUDE_DIRS}")
endif()
```

could be changed by the below with keeping https://github.com/3dem/relion/commit/6918e6d945b618d86314b8dbb339cc240b3c8861.

```cmake
if (NOT "${MPI_CXX_INCLUDE_PATH}" STREQUAL "")
        include_directories("${MPI_CXX_INCLUDE_PATH}")
elseif (NOT "${MPI_CXX_INCLUDE_DIRS}" STREQUAL "")
        include_directories("${MPI_CXX_INCLUDE_DIRS}")
endif()
```

The original cmake configuration will produce error since MPI_CXX_INCLUDE_DIRS/MPI_CXX_INCLUDE_PATH is empty if mpiicc/mpiicpc is used as C/C++ compiler command and it is true since mpiicc and mpiicpc does not need to include specific MPI header files.

tjragan:
@CharlesCongdon's suggestion worked.  I was using `CC=mpiicc CXX=mpiicpc cmake3 -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc` as suggested by the icc docs, but `cmake3 -DCMAKE_C_COMPILER=icc -DCMAKE_CXX_COMPILER=icpc -DMPI_C_COMPILER=mpiicc -DMPI_CXX_COMPILER=mpiicpc` has fixed it.  Thanks!

PR #767, opened 2021-04-07
Fixed buffer overflow in FileName::compose

andschenk:
The temporary aux_str buffer  in Filename::compose(long int, const std::string &, int) needs to have a size of numberlength+2 (length of number + @ character + 0 terminator) instead of numberlength+1 to prevent it from overflowing in the sprintf function.


biochem-fan:
Great discovery! Thank you very much. I merged this to `ver3.1` branch.

PR #783, opened 2021-06-30
Support for disabling and tuning stream buffers

andschenk:
This patch adds support for tuning the size of the stream buffer for reading mrcs stacks in Relion.

Rationale for the patch:
When running Relion on large, network attached file systems it can show a very unfavorable caching behavior when reading particle stacks. With large Lustre and GPFS file systems we have seen that data requested through the network (IB in our case) can reach 10x the data actually read by Relion. This can put a large strain on network and storage in cluster settings, where multiple nodes are connected to the same network storage and can lead to a degradation of processing performance in Relion due to I/O waits. 
The issue is probably caused by the fact that the data read from the filesystem is cached twice, first by the filesystem cache and then second by the default stream buffer created by the fopen() calls used in Relion. If the stream buffer is set too big by default and a lot of small random I/O operations are performed (e.g. during reading of particles in random order), this can lead to cache trashing and re-reading of the data from the network storage.

With the patch the environment variable RELION_STACK_BUFFER can be set with export/setenv to adjust the buffering behavior:

RELION_STACK_BUFFER not defined: Relion uses the default stream buffer size for reading mrcs stacks (same behavior as before the patch)
RELION_STACK_BUFFER set to 0: stream buffering is disabled for reading mrcs stacks
RELION_STACK_BUFFER set to X: stream buffer size is set to X bytes for reading mrcs stacks

After disabling the stream buffer for reading stacks we saw a tremendous drop in network traffic and better caching behavior. I.e. the network traffic to the storage server for the test job pretty much dropped to 0 after the initial noise estimate, as particle data was read from cache during iterations.

### DIFF
--- a/src/image.h
+++ b/src/image.h
@@ -227,7 +227,15 @@ public:
 
 	void openFile(const FileName &name, int mode = WRITE_READONLY)
 	{
-
+		static bool buffSet=false;
+		static size_t buffSize=SIZE_MAX;
+		if(!buffSet){
+			char * buffStr = getenv("RELION_STACK_BUFFER");
+			if(buffStr!=NULL){
+				buffSize=std::atoi(buffStr);
+			}
+			buffSet=true;
+		}
 		// Close any file that was left open in this handler
 		if (!(fimg ==NULL && fhed == NULL))
 			closeFile();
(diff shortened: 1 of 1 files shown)


biochem-fan:
Thank you ver much for your suggestion.

This looks good to me but I hesitate to merge this into 3.1, which is being stabilized.
I will merge this to 4.0, which will be made public soon for beta testing.

Also out of curiosity: do you have quantitative data or plots to show the effect of this?
e.g. bandwidth usage, IOPS, time for a typical run?

biochem-fan:
I want to remove `static` to make this function thread-safe. Can you test if calling `getenv` many times does not degrade performance?

zao:
> I want to remove `static` to make this function thread-safe; multiple threads can enter the `if` block simultaneously

You could replace it with C++11 thread-safe statics - this initialization using an immediately invoked lambda function will be guarded by the compiler:
```cpp
static const size_t buffSize = []() -> size_t {
    char * buffStr = getenv("RELION_STACK_BUFFER");
    if(buffStr!=NULL){
        return std::atoi(buffStr);
    }
    return SIZE_MAX;
}();
```

biochem-fan:
@zao Thanks! This is very cool. I didn't know this trick.

biochem-fan:
I committed this to 4.0 (not public yet) in 3d89b6c6. Thank you very much for your contribution.

andschenk:
@zao : Thanks for this trick. This makes the code much cleaner.

PR #788, opened 2021-08-03
Fix for race condition during cleanup of local scratch.

andschenk:
After a classification/refinement using a scratch directory is finished, each worker rank tries to delete the scratch directory.
This can lead to a race condition, where some of the rm -fr commands fail and where Relion aborts and marks the job as failed.
Setting keep_scratch=true for all ranks except the first on each node avoids the race condition, as only one rank per node performs the cleanup.


biochem-fan:
This makes sense. Thanks for your contribution.

PR #793, opened 2021-09-03
Enabled FLTK_SKIP_FLUID in CMakeLists.txt - Do not check for fltk fluid

pmargara:
Dear All,
on some distributions like RH derivates, when fltk is coming from repos, you get errors related to the absence of the command fluid in FLTK that require manual modifications in some files generated by cmake. If the FLTK_SKIP_FLUID parameter is appropriately configured in the CMakeLists.txt file, the errors found disappear. Since from my tests the presence of fluid is not binding on the correct functioning of the relion gui, I propose the following modification. 

Greetings,
    Paolo


biochem-fan:
This looks fine. I merged it into the ver3.1 branch. Thank you very much for your suggestion.

PR #800, opened 2021-09-13
changed sign of twist corr

### LINKED ISSUE #799: helical_nstart option for refinement is broken
The `--helical_nstart` option of `relion_refine` does not work in the current main version - instead of reverting the twist back to the n-start convention, it adds a multiple of 360 on top (e.g. a 2-start helix with 180.5° 1-start twist becomes 721° instead of 1° in 2-start convention). For my application - a C2_1 amyloid, where rot priors don't work out of the box - using `--helical_nstart=2` hence made matters even worse. The problem results from a typo in `helix.cpp` (line 4425). After fixing it, it works well (at least for the C2_1 amyloid in both right-handed and left-handed config) and improves the reconstruction significantly.

robertbuecker:
fixes #799
Tested with left- and right-handed 2-start helical reconstructions.

### DIFF
--- a/src/helix.cpp
+++ b/src/helix.cpp
@@ -4422,7 +4422,7 @@ void updatePriorsForHelicalReconstruction(
 		{
 			helical_rise[iclass] *= helical_nstart;
 			RFLOAT aux = helical_twist[iclass] * helical_nstart;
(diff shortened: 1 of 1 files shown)


biochem-fan:
@sjors approved this. I will merge it. Thank you very much for your contribution.

PR #801, opened 2021-09-13
Allows to compile without MPI

benoistlaurent:
This feature has been proven useful when installing relion on a cluster master node, where MPI libraries are not always available.


biochem-fan:
Thank you very much of your suggestion but I am not sure if such a use case is common enough to justify inclusion of this patch. In general we don't want to have many `ifdef`s into our code. For example, we made libtiff mandatory and removed a lot of `ifdef`s recently, because one can build and install it without root privileges. The same holds for OpenMPI. The only exception is GUI, since installing FLTK and all X11 dependencies is very tricky.

biochem-fan:
My point is whether the added complexity, not only `ifdef`, is worth maintaining. For example, if Jasenko adds another program, we must update the list of `list(FILTER REL_JAZ_XXX EXCLUDE REGEX ".*_mpi.h")`s. We often forget it and such a build will quickly get broken, especially for features we ourselves never use and test.

biochem-fan:
This introduced a lot of code duplication. I myself is against this PR, but if other members decide to merge it, the first state is better.

benoistlaurent:
Well, at some point one have to choose: it's either `ifdef` either code duplication...
I'm not a big fan of code duplication myself so I can revert back to `ifdef` if you'd like...

biochem-fan:
The main point is that I don't think it is unreasonable to request users to install OpenMPI themselves.

benoistlaurent:
I would agree in most cases. As stated in the MR description, I developed this feature for my personal needs, i.e. installing the GUI on a cluster master node, where MPI isn't and won't be installed for technical reasons. I would imagine I'm not the only one to be in such situation.

PR #805, opened 2021-09-24
Ver4.0 Install on Centos

schiracha:
Installing Dependencies for Centos 7 and 8 (RedHat), the libraries for C compiler, tiff and fftw have slightly different repositories.  "cuda" can also be added to this list for those running nvidia style cards.
sudo yum install cmake git gcc gcc-c++ openmpi-devel libtiff-devel fftw-devel libpng-devel


biochem-fan:
Hi, why does your diff have so many commits? Please base your changes on our ver4.0 branch.

biochem-fan:
Thanks. I added instructions for Redhat like systems in README of the `ver4.0` branch.

https://github.com/3dem/relion/blob/ver4.0/README.md

PR #811, opened 2021-09-28
Refactor usage of which

aohnsc:
In our environment it is not possible to use the `which` command,
hence we need to patch pipeline_jobs.cpp before we are able to compile
and use RELION.

Currently `which <cmd>` is used in several places in pipeline_jobs.cpp which
makes the patching a bit tedious. This commit adress this problem by
introducing a helper function obtainFullPathForCommand(), which converts
a string "cmd" into "`which cmd`". By using the helper function, it
means that we only need to change one row in this file before we are
able to build and use it.

### DIFF
--- a/src/pipeline_jobs.cpp
+++ b/src/pipeline_jobs.cpp
@@ -1417,9 +1417,9 @@ bool RelionJob::getCommandsMotioncorrJob(std::string &outputname, std::vector<st
 
 	std::string command;
 	if (joboptions["nr_mpi"].getNumber(error_message) > 1)
-		command="`which relion_run_motioncorr_mpi`";
+		command = obtainFullPathForCommand("relion_run_motioncorr_mpi");
 	else
-		command="`which relion_run_motioncorr`";
+		command = obtainFullPathForCommand("relion_run_motioncorr");
 	if (error_message != "") return false;
 
 	// I/O
@@ -1656,9 +1656,9 @@ bool RelionJob::getCommandsCtffindJob(std::string &outputname, std::vector<std::
 	inputNodes.push_back(node2);
 
 	if (joboptions["nr_mpi"].getNumber(error_message) > 1)
-		command="`which relion_run_ctffind_mpi`";
+		command = obtainFullPathForCommand("relion_run_ctffind_mpi");
 	else
-		command="`which relion_run_ctffind`";
+		command = obtainFullPathForCommand("relion_run_ctffind");
 	if (error_message != "") return false;
 
 	command += " --i " + joboptions["input_star_mics"].getString();
@@ -1786,7 +1786,7 @@ bool RelionJob::getCommandsManualpickJob(std::string &outputname, std::vector<st
 	initialisePipeline(outputname, job_counter);
 	std::string command;
 
-	command="`which relion_manualpick`";
+	command = obtainFullPathForCommand("relion_manualpick");
 
 	if (joboptions["fn_in"].getString() == "")
 	{
@@ -1969,7 +1969,7 @@ bool RelionJob::getCommandsAutopickJob(std::string &outputname, std::vector<std:
 
 		label += ".continuemanual";
 
-		command="`which relion_manualpick`";
+		command = obtainFullPathForCommand("relion_manualpick");
 
 		command += " --i " + joboptions["fn_input_autopick"].getString();
 		command += " --odir " + outputname;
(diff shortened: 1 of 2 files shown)


biochem-fan:
Why you cannot use `which`? Cannot you workaround this by defining appropriate shell aliases or writing a wrapper script?

aohnsc:
We are controlling where certain programs are installed in the system explicitly. We modify the PATH environment variable accordingly in order to direct the program where the other programs are installed. This pattern renders the call to `which` superfluous. Basically we want to avoid the extra fork it takes to do `which`.

Note that the commit doesn't change the current behavior of pipline_jobs.cpp.

biochem-fan:
> We are controlling where certain programs are installed in the system explicitly. We modify the PATH environment variable > accordingly in order to direct the program where the other programs are installed.

Doesn't this mean you set `PATH` to RELION installation? Why don't you use a more standard way, e.g. `module`?

Although your patch does not changes the program behaviour, we are very hesitant to modify our program only to support a very idiosyncratic system configuration.

>  Cannot you workaround this by defining appropriate shell aliases or writing a wrapper script?

Doesn't this work?

aohnsc:
> > We are controlling where certain programs are installed in the system explicitly. We modify the PATH environment variable accordingly in order to direct the program where the other programs are installed.
> 
> Doesn't this mean you set `PATH` to RELION installation? Why don't you use a more standard way, e.g. `module`?

We are using module, and module is modifying the PATH environment variable:

```
$ echo $PATH | wc
      1       1     190
$ module load RELION
$ echo $PATH | wc
      1       1    4352
```

> Although your patch does not change the program behaviour, we are very hesitant to modify our program only to support a very idiosyncratic system configuration.

I understand this and it makes sense to me if you are reluctant to change the program, but at the same time I want you to understand that this change will simplify my job very much and your code will be more decoupled.

> > Cannot you workaround this by defining appropriate shell aliases or writing a wrapper script?
> 
> Doesn't this work?

In our compute cluster we have approximately 1000 users and I have no statistics of how many that are using Relion regularly. But it does not make sense to introduce Relion specific idiosyncratic system configuration for all our users.

I'm not entirely sure how a shell alias or wrapper script to `which` will help us anyway.

src/pipeline_jobs.cpp
```diff
@@ -6641,3 +6641,7 @@ bool RelionJob::getCommandsTomoReconPartJob(std::string &outputname, std::vector
 
 }
 
+std::string RelionJob::obtainFullPathForCommand(const std::string &cmd)
+{
+	return "`which "+ cmd +"`";
```
aohnsc, line 6646: This will be the only row that I need to change when I create a new build of Relion (instead of changing 49 rows).

biochem-fan:
> We are using module, and module is modifying the PATH environment variable:

OK, then RELION should work as is, right?

Users who want to use RELION will invoke `module load RELION in `.bashrc` or the job template,
so `which relion_refine` executed by RELION will find the right binary, won't it?

I don't understand why you have to change RELION itself.

biochem-fan:
I re-read your earlier comments.

You wrote:
> This pattern renders the call to which superfluous. Basically we want to avoid the extra fork it takes to do which.

So the presence of `which` does not break anything, right? It is only **ONE** fork per job. I don't believe this causes noticeable performance degradation and is worth the hassle of removing it.

aohnsc:
Thanks for your replies.

I have difficulties understanding why you need to call e.g. `which relion_refine` in the first place. Can you please elaborate on that?

aohnsc:
> Can you also reply to my questions?

Of course!

> > We are using module, and module is modifying the PATH environment variable:

> OK, then RELION should work as is, right?

> So the presence of which does not break anything, right?

I'll get back to you on these questions the next time I'm building Relion.

> so `which relion_refine` executed by RELION will find the right binary, won't it?

Yes, `which relion_refine` will find the right binary and show the full path to it. The which command will use the PATH environment variable and pick the first match for "relion_refine", in the same way as the shell uses PATH.

> Indeed which is not necessary for non-MPI jobs. For MPI jobs, some MPI runtimes need a full path.

Do you have a source to back that some MPI runtimes need a full path? (just for my clarity)

It doesn't seem that you are very eager to do add this proposed change and I'm considering to close the pull request if you don't mind.

biochem-fan:
OK, I am closing this issue.

> Do you have a source to back that some MPI runtimes need a full path? (just for my clarity)

I don't remember. This was necessary many years ago. Modern MPI runtime might not have this issue.

In future, CCPEM, instead of RELION, will handle GUI, jobs and command line generation. At that point, this issues will be refactored out.

Issue #812, opened 2021-09-30
"make install" makes two copies of some programs with inconsistent rpaths


colinpalmer:
I'm trying to install RELION 4 outside the build directory by setting `CMAKE_INSTALL_PREFIX` and using `make install`. This currently (version 4.0-beta-1-0820b899) creates two copies of all the `exp` and `tomo` and `spa` programs, one of which has the install rpath set correctly and one of which doesn't.

I'd be happy to help fix this issue but first I need to know what the correct structure is intended to be.

Some details:

- Lines in `src/apps/CMakeLists.txt` set the runtime destination for [`spa`](https://github.com/3dem/relion/blob/0820b89905bd7308e7d2486cdcf98c8eea7fc7a0/src/apps/CMakeLists.txt#L400), [`tomo`](https://github.com/3dem/relion/blob/0820b89905bd7308e7d2486cdcf98c8eea7fc7a0/src/apps/CMakeLists.txt#L417) and [`exp`](https://github.com/3dem/relion/blob/0820b89905bd7308e7d2486cdcf98c8eea7fc7a0/src/apps/CMakeLists.txt#L437) targets to `bin/spa`, `bin/tomo` and `bin/scripts` directories respectively.
- [These lines](https://github.com/3dem/relion/blob/0820b89905bd7308e7d2486cdcf98c8eea7fc7a0/CMakeLists.txt#L405-L408) in the top-level `CMakeLists.txt` install all files from `build/bin` into `CMAKE_INSTALL_PREFIX/bin`.
- During `make`, all executables are built in the `build/bin` directory, regardless of their runtime install destination.
- During `make install`, the `spa`, `tomo` and `exp` programs are installed twice: once as executable targets, meaning they go into their respective `bin/spa`, `bin/tomo` and `bin/scripts` directories with their rpaths set correctly; and once as simple files, meaning they're copied directly from `build/bin` to the install `bin` directory with their rpaths unchanged.

It looks like the intention is that these programs should end up in subdirectories of `bin`, but that means people who use `make install` will need to add multiple directories to their `PATH`, and the layout of executables in the build and install locations will be inconsistent.

I'd suggest the simplest way to fix this is to remove the subdirectories completely, so the `spa`, `tomo` and `exp` programs have runtime destination `bin` like everything else. Then, the build and install layouts will be identical and the rpaths should be set correctly. Is that acceptable?

[...]

PR #813, opened 2021-09-30
Install all executables in bin directory

colinpalmer:
This avoids use of subdirectories, which require a more complicated `PATH` and can cause inconsistencies in build and installation.

https://github.com/3dem/relion/issues/812


biochem-fan:
Thank you very much for your quick fix!

Issue #896, opened 2022-07-13
[Question] Is Linux ARM64 supported ?


julien-faye:
Hello RELION community.

I’d like to ask whether Linux ARM64 is an officially supported platform? For hardware like AWS Graviton2 / Ampere ?
Do I need to do something special to be able to build and run RELION on ARM64 ?

I searched around but I was not able to find any related information.

Thank you!

biochem-fan:
ARM64 is not officially supported. It should work (albeit slower), but we cannot test or debug. We don't have access to any ARM64 platform.

julien-faye:
Thanks for the quick answer, @biochem-fan !

I have seen OSS projects applying at https://github.com/WorksOnArm/equinix-metal-arm64-cluster for hardware that could be used for testing, debugging, CI purposes!
To apply one just have to open a Github issue.
I hope you consider it!

martin-g:
Hi,

Huawei Cloud is also willing to donate an ARM64 VM for long term usage with no strings attached!

dzyla:
Hi,
I could compile Relion4 on Raspberry Pi, but it is hard to tell if all options work due to limited RAM and CPU. The compilation process works tho.

julien-faye:
Thanks for sharing this info, @dzyla !
I'm glad there are more people interested in ARM64!

rbs-sci:
Around three years ago, I got RELION 3 to compile and everything would work on a 4GB nVidia Jetson Nano (I ran through the cisTEM tutorial dataset as it's small) and the only issue is CTFFIND... which works, provided you manually replace some of the x86 assembly used for speed with platform agnostic C code.

The only caveats were the limited memory (I ended up making a big swap file on a USB3 connected SSD or polishing would crash) and obviously how slow it was (despite CUDA acceleration). On the newer nVidia ARM64 development systems it'd probably do much better... but they are _way_ out of "tinkering" budget.

I'll dig the info out from my archives if anyone is interested.

martin-g:
@rbs-sci Not sure you noticed the Pull Request lined above. It added Github Actions based testing on x86_64 and aarch64 - https://github.com/3dem/relion/blob/master/.github/workflows/build.yml

PR #902, opened 2022-08-10
Add CI based on Github Actions

martin-g:
It runs on Github-hosted runner natively for Linux x86_64 and in a
emulated (QEMU) Docker container for Linux aarch64

Rlated-to: #896

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,37 @@
+on: [ pull_request ]
+
+name: Build RELION
+
+jobs:
+  build_on_x86_64:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt-get update -q -y && \
+          sudo apt-get install -q -y cmake git build-essential mpi-default-bin mpi-default-dev libfftw3-dev libtiff-dev libfltk1.3-dev
+      - name: Build RELION for Linux x86_64
+        run: mkdir build && cd build && cmake .. && make
+
+  build_on_aarch64:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Build RELION for Linux aarch64
+        uses: uraimo/run-on-arch-action@v2
+        with:
+          arch: aarch64
+          distro: ubuntu20.04
+          githubToken: ${{ github.token }}
+          dockerRunArgs: |
+            --volume "${PWD}:/relion" 
+          install: |
+            apt-get update -q -y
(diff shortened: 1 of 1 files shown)


martin-g:
Any feedback on this PR ?

biochem-fan:
We have zero experiences with Github Actions (and "CI" concepts in general)...

martin-g:
@biochem-fan I'd be happy to help and teach you if you'd like!
The idea is to setup Github Actions as CI and every time you receive a Pull Request  the CI will execute the build steps and report if the build fails for any reason. This way the contributor will know that his/her changes cause regression and will have to fix it or the PR won't be accepted.

The CI config in this PR is rather simple and I don't expect that it will become more complex in the future.
It just installs the needed prerequisites and runs `make`.
There is a slight duplication because at the moment Github Actions supports only x86_64 VMs and to run the build on aarch64 one has to use QEMU via uraimo/run-on-arch-action action.

biochem-fan:
Do we have to do something else to enable this in addition to merging this pull request?

https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration says "GitHub Actions usage is free for standard GitHub-hosted runners in public repositories". So we won't get billed, will we? Is it free even when there are many pull requests and commits and the job is executed many times?

biochem-fan:
Can we also test CUDA builds and accelerated CPU builds (`cmake -DALTCPU=ON`)?

martin-g:
> So we won't get billed, will we? Is it free even when there are many pull requests and commits and the job is executed many times?

It is completely free!
The CI workflow will be executed whenever a Pull Request is created and then after each additional commit to the branch for this PR.
There is nothing else to do from your side! 

At the moment the CI workflow is configured to execute only for Pull Requests. I could also add logic to be executed on push to `master` branch. This would be useful if the project members/committers do not use Pull Requests.

martin-g:
> Can we also test CUDA builds and accelerated CPU builds (`cmake -DALTCPU=ON`)?

I haven't used those myself in other projects but it seems you can:
* https://github.com/marketplace/actions/cuda-toolkit
* https://github.com/ptheywood/cuda-cmake-github-actions/blob/master/.github/workflows/Ubuntu.yml

biochem-fan:
Thanks for clarification. I will merge this PR to see what happens. If you have time, can you try our accelerated CPU build in another pull request? The accelerated CPU code path does not need CUDA. Just add `-DALTCPU=ON` to `cmake`. This will also test whether this CI hook is working.

>  I could also add logic to be executed on push to master branch. This would be useful if the project members/committers do not use Pull Requests.

The cutting-edge development is happening on the other private repo, so this is not necessary to this public repo.

biochem-fan:
OK, now I see "Build RELION" on the "Actions" tab.

martin-g:
I will send the new PR in few minutes!

PR #904, opened 2022-08-16
Fix for LLVM-based compiler

do-jason:
The existing "#pragma omp simd" in diff2_coarse function in CPU accelerated code is wrong since it requires private copy of partial array(diff2s) for SIMD vectorization. It is not supported by OpenMP standard. And this is causing incorrect result on LLVM-based compilers(Clang, Intel oneAPI, AMD AOCC, ARM clang, ...) while Intel and GCC compilers are okay.

And there is one fix for CMakeList.txt. It will fix Intel oneAPI compiler MKL support and this change is already applied "relion-devel" repo.


biochem-fan:
Is this `qmkl` flag supported by not-the-latest ICC as well?

do-jason:
@biochem-fan Intel calls icc/icpc as classic compiler and icx/icpx as oneAPI compiler. And this is for oneAPI compiler. oneAPI compiler has added warning in some versions if "-mkl" is used but "-qmkl" will always work for oneAPI compiler versions so far.

biochem-fan:
Is this compatible with old (say 2018) Intel classic compiler?

do-jason:
@biochem-fan This will only change for oneAPI compiler, and it remains the same for old Intel compiler.

biochem-fan:
OK, I merged this.

Issue #906, opened 2022-08-22
Building with CMake newer than 3.9.6 does not work


clil16:
Using a CMake version greater than CMake v3.9.6 results in an error.  My guess is that it has to do with the changes made to the FindMPI module when CMake version 3.10.0 was released.  I can confirm the issue is present in version 3.10.0, 3.18.2, and 3.22.2 of CMake.

To clarify:
CMake 3.9.6 - works
CMake 3.10.0 - does not work
CMake 3.18.2 - does not work
CMake 3.22.2 - does not work

**Environment:**
 - OS: Red Hat Enterprise Linux release 8.4
 - Compiler: Intel Compiler 2021.5.0 20211109
 - MPI runtime: OpenMPI 4.1.4
 - RELION version: RELION ver4.0 branch commit 2dffb8415a7e09f14532668b06e353fad1c46b2a
 - CMake version: 3.10.0+
 - Memory: 196 GB
 - GPU: None

**CMake command:**
 - Full command:
   ```
   cmake -DCMAKE_C_COMPILER=mpicc -DCMAKE_CXX_COMPILER=mpicxx -DALTCPU=ON -DMKLFFT=ON -DGUI=ON -DCUDA=OFF -DCudaTexture=OFF -DFORCE_OWN_FLTK=ON -DCMAKE_C_FLAGS="-O3 -ip -g -xCOMMON-AVX512 -restrict" -DCMAKE_CXX_FLAGS="-O3 -ip -g -xCOMMON-AVX512 -restrict" -DCMAKE_BUILD_TYPE=Release -DTIFF_INCLUDE_DIR=/sw/pkgs/coe/o/image-libraries/220318/include -DTIFF_LIBRARY=/sw/pkgs/coe/o/image-libraries/220318/lib/libtiff.so -DPNG_INCLUDE_DIR=/sw/pkgs/coe/o/image-libraries/220318/include -DPNG_PNG_INCLUDE_DIR=/sw/pkgs/coe/o/image-libraries/220318/include -DPNG_LIBRARY=/sw/pkgs/coe/o/image-libraries/220318/lib/libpng16.so -DJPEG_INCLUDE_DIR=/sw/pkgs/coe/o/image-libraries/220318/include -DJPEG_LIBRARY_RELEASE=/sw/pkgs/coe/o/image-libraries/220318/lib/libjpeg.so -DCMAKE_INSTALL_PREFIX=/sw/pkgs/lsi/relion-cpu/4.0 ..
   ```

[...]

PR #907, opened 2022-08-25
Check additional MPI header's emptiness

do-jason:
Fix for https://github.com/3dem/relion/issues/906


biochem-fan:
This was merged in relion-devel.

PR #914, opened 2022-09-09
Build with/without ALTCPU

martin-g:
Add build option -DALTCPU as requested by @biochem-fan at https://github.com/3dem/relion/pull/902#issuecomment-1241586402

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,9 @@ name: Build RELION
 jobs:
   build_on_x86_64:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        altcpu: [ON, OFF]
     steps:
       - uses: actions/checkout@v3
       - name: Install dependencies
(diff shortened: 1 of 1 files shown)


biochem-fan:
Great!

It is interesting that this pull request already runs ALTCPU=ON tests, even before I merge itself.

martin-g:
Yes, that's the idea - the suggested changes by a contributor to be tested before being merged to `master`.

biochem-fan:
> the suggested changes by a contributor to be tested before being merged to master.

But aren't the tests defined by the repository (in this case your first PR already-merged), not the (second) pull request itself?

martin-g:
The first PR was a bit special - it activates the Github Actions. Once activated any kind of changes (in .cpp files or in the workflow itself) are tested.

biochem-fan:
OK.

martin-g:
All checks have passed! 
The aarch64 ones are slower due to the emulation. Microsoft Azure started providing ARM64 VMs since Sep 1, so hopefully soon Github will start using them too and then there won't be a need of emulation!

biochem-fan:
Wonderful. I merged it.

PR #920, opened 2022-09-29
This codes should be in common code block not CUDA specific

do-jason:
This is simple fix and the title explains the bug.


biochem-fan:
Thank you very much. This makes sense.

PR #926, opened 2022-10-12
Fix for projectors[].clear() loop

do-jason:
Loop of clearing projector class fix.


biochem-fan:
Great find!

Do you mean the number of projectors and that of backprojectors are different in MultiBody refinement?

biochem-fan:
I merged your PR. Thank you very much!

PR #979, opened 2023-07-13
make EER available

martinschorb:
It was not possible to call MotionCorr with EER files yet. This change should make the correct parameter available.


martinschorb:
It would be very neat if that's all what's needed...

Since I have no experience in what else happens here, please feel free to dump this suggestion. 

I just stumbled across MotionCor2 crashing without explicitly pointing out the file type being the problem and found that calling it with `-Eer` instead of `-Mrc` makes it work.

biochem-fan:
This lets you run UCSF MotionCor2 on EER movies but won't let you perform Bayesian Polishing afterwards.
To support the latter, more complicated changes are necessary but I have no time and intention to work on it.

PR #1015, opened 2023-11-01
Change MPI block size and minor SYCL document update

do-jason:
This will update default MPI block size only when "-DUSE_MPI_COLLECTIVE" is enabled. This is default for ALTCPU and SYCL. It has no impact when "-DUSE_MPI_COLLECTIVE" is not set.
The default 512MB for MPI collective may increase file descriptor usage and result in error or lower performance at multi-node scale.

I've erroneously committed in "relion-devel" branch without PR and there is closed PR to wrong "master" branch", sorry about that.


biochem-fan:
Thanks. This looks OK to me.

Issue #1017, opened 2023-11-01
"ACC_PTR_DEBUG_INFO" is undefined 


james-vincent:
**Describe your problem**

Compilation fails on:

`src/fftw.h(473): error: identifier "ACC_PTR_DEBUG_INFO" is undefined`

- ACC_PTR_DEBUG_INFO macro is used in fftw.h under #ifdef DEBUG_CUDA

 - ACC_PTR_DEBUG_INFO is defined in src/acc/acc_ptr.h

  - acc_ptr.h is not  #include  in fftw.h

**Environment:**
Building with spack

 - OS: CentOS 7.9.2009
   gcc@8.4.0 
   cuda@11.6.2 
   openmpi@3.1.6

biochem-fan:
Why did you enable `DEBUG_CUDA`? This flag is not relevant to end users.

james-vincent:
I did not enable, on purpose.

I’ve found the flag now and can see it was enabled for nvcc:

This is the call. I’ve snipped the parent paths to make it cleaner to read:

cuda/11.6.2/bin/nvcc <path>/src/acc/cuda/cuda_helper_functions.cu -c -o <path>/acc/cuda/./relion_gpu_util_generated_cuda_helper_functions.cu.o -ccbin

 /programs/x86_64-linux/gcc/8.4.0/bin/g++ -m6
            4 -DINSTALL_LIBRARY_DIR=<path>/lib
            -DSOURCE_DIR=<path>src/
           -DACC_HIP=3 -DACC_CUDA=2 -DACC_CPU=1 -D_CUDA_ENABLED -DHAVE_SINCOS -DHAVE_TIFF -DHAVE_PNG -DHAVE_JPEG -Xcomp
            iler ,\"-g\",\"-O2\",\"-g\",\"-DNDEBUG\",\"-O2\",\"-DDEBUG\",\"-DDEBUG_CUDA\" -DNVCC


Seems like a hidden setting or something on our end.


James Vincent, PhD
Bioinformatics Software Curator
Dept. of BCMP, Harvard Medical School
— BioGrids.org --

james-vincent:
Upon further inspection, the release with debug on is the default in the relion spack package we have used in the past.  This did not cause an issue in 4.0.1:

./cmake/BuildTypes.cmake:set(RELION_DEFINITIONS_RELWITHDEBINFO "-DDEBUG_CUDA")

The debug statements in fftw.h aere new to the 5.0 beta release:

[...]

PR #1021, opened 2023-11-02
Fix for https://github.com/3dem/relion/issues/1017

do-jason:


biochem-fan:
Thank you very much for your prompt fix.

PR #1028, opened 2023-11-10
ManualPicker: automatically save when closing the window

biochem-fan:
ManualPicker says:

> NOTE: in order to write the new list of coordinate STAR files, you need to re-count the particles or quite this program through the File menu. Do NOT kill the program using the operating system's window manager!

but this is not user-friendly. Most people don't notice this message, and even if they do see this, they just forget to use the menu.

With this patch, ManualPicker automatically saves the list even when closed by the window manager.

This also fixes a problem in the `[File]-[Quit]` menu. It used to save the list *without* recounting. Now it recounts and then saves the list.

### DIFF
--- a/src/displayer.cpp
+++ b/src/displayer.cpp
@@ -1982,7 +1982,10 @@ int pickerViewerCanvas::handle(int ev)
 			else if ( strcmp(m->label(), "Help") == 0 )
 				printHelp();
 			else if ( strcmp(m->label(), "Quit (CTRL-q)") == 0 )
-				exit(0);
+                        {
+                            saveCoordinates(false);
+                            exit(0);
+                        }
 			redraw();
 			return 1; // (tells caller we handled this event)
 		}
--- a/src/manualpicker.cpp
+++ b/src/manualpicker.cpp
@@ -582,7 +582,7 @@ void manualpickerGuiWindow::cb_menubar_quit(Fl_Widget* w, void* v)
 void manualpickerGuiWindow::cb_menubar_quit_i()
 {
 	cb_menubar_recount_i();
-	exit(0);
+	RELION_EXIT_SUCCESS;
 }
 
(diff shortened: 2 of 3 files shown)


scheres:
Hi Takanori, 
Thanks for this suggestion, but could it be that this does not work? If I close the window from the desktop manager, the coordinates do still not get saved? Am I doing something wrong, and does this work on your system?
Sjors

biochem-fan:
@scheres Let's make sure we are looking at the same thing.

> If I close the window from the desktop manager

I mean closing the micrograph list window (`relion_manualpick`) with the top right [X] button of the title bar. Are you closing the window in other ways?

> the coordinates do still not get saved?

This patch is about automatically recounting and updating `manualpick.star`, which contains the list of micrographs with picked particles, upon closing `relion_manualpick`. This does not automatically save picked coordinates in individual micrograph view (`relion_display`).

biochem-fan:
@scheres Do we want to auto-save picking results on individual micrographs (i.e. `relion_display`)?
If so, you can use a closing callback as in dd7d0561.

scheres:
That would be nice. I tried this actually yesterday, but could not make it work, as the picker uses a separated canvas. Perhaps you're better at this. But it is not too high on the priority list... People rarely pick manually these days.

biochem-fan:
It is hard to force `relion_manualpick` to recount particles when new coordinates are saved by `relion_display`. This needs an `inotify` mechanism. On the other hand, it is easy to auto-save coordinates in `relion_display`. However, I am not sure if this is desirable. People might accidentally click on a micrograph when inspecting AutoPick results. Such changes are discarded unless users explicitly save them. Do we want to automatically save instead?

scheres:
OK, let's just leave it as it is. Thanks!

biochem-fan:
If we decide so, perhaps your commit b8cbd4f0 should be reverted for consistency.

scheres:
ok, done.

PR #1039, opened 2023-11-30
Implemented memory pool for SYCL device/host memory

do-jason:
Implemented memory pool for SYCL device/host memory. (added src/acc/sycl/sycl_memory_pool.*)
- All except ZE_* and ZEX_* environment variables are not necessary.
- Now "ulimit -n" is not necessary in most cases.

Tuned SYCL kernel workgroup size.
- Showed significant improvement in 2D classification performance.

Fixed AccPtr host memory leak in PassWeights.weights.  (src/acc/acc_ml_optimiser_impl.h)
Prevented threads of the same MPI rank from being assigned to different SYCL devices in automatic mapping (ml_optimiser_mpi.cpp and ml_optimiser.cpp line 1740)
Fix Intel IPP library header file for oneAPI 2024 release and older. (CMakeLists.txt and src/acc/utilities.h)
Minor update for SYCL2020 standard.
Consistent use of tab/space in source codes under src/acc/sycl directory.


biochem-fan:
Thank you very much for your continuing contributions!

I am bit busy right now (MicroED and XFEL beam time) but hopefully will try to compile this by next Wednesday. (or can @scheres test this when you recompile the LMB binary?)

By the way, please don't squash your commits when sending your PR next time so that I can see which change corresponds to which goal.

PR #1049, opened 2023-12-15
Patch for DoublePrec_ACC build using SYCL

do-jason:
This will add support of double precision build for SYCL-acceleration.


biochem-fan:
Thank you very much. This looks OK to me.

Issue #1016, opened 2023-11-01
Support CMake CMP0146 (i.e. FindCUDA module removal)


biochem-fan:
We have to support this, otherwise new versions of CMake will fail.
https://cmake.org/cmake/help/latest/policy/CMP0146.html

biochem-fan:
Someone using NERSC faced this issue, as [reported on CCPEM](https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind2402&L=CCPEM&O=D&P=27974).

@scheres Do you think it is reasonable to drop support for CMake < 3.09 in RELION 5.0?

biochem-fan:
I hope [the cmake_no_findCUDA branch](https://github.com/3dem/relion/tree/cmake_no_findCUDA) branch (after commit c900cf) should work. Please test this if you have time.

green-br:
Any update on this change- should we try and test the branch or will this be included soon?

biochem-fan:
@green-br The effort has been kind-of stalled; if you can provide feedbacks (if compilation succeeds and if computational results are resonable), it would be very useful.

PR #1077, opened 2024-02-05
Update CMakeLists to support newer CMake versions

Qiangong2:
FIND_PACKAGE(CUDA) was deprecated starting with cmake 3.10 and flat out removed in cmake 3.27. For machines in environments with strict update requirements, this breaks compilation on newer systems.

FIND_PACKAGE(CUDAToolkit) works the exact same way, and is supported by cmake.

Thanks to the NERSC team for identifying the issue.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -136,10 +136,11 @@ endif()
  #This *has* to be AFTER project()
 include(${CMAKE_SOURCE_DIR}/cmake/BuildTypes.cmake)
 
-if(CUDA)
-    # -----------------------------------------------------------------------------CUDA--
-    # DOC: https://cmake.org/cmake/help/v3.28/module/FindCUDA.html
-    FIND_PACKAGE(CUDAToolkit)
+if(CMAKE_VERSION VERSION_LESS 3.10)
+	FIND_PACKAGE(CUDA)
+else(CMAKE_VERSION)
+	# newer version of CMake use a different module name
+	FIND_PACKAGE(CUDAToolkit)
 endif()
 
 if(CUDA_FOUND)


biochem-fan:
Thank you very much but can you make the pull request compatible with older versions of CMake?

Qiangong2:
CUDAToolkit was added in 3.10, and distributions as far back as RHEL/CentOS 7 all have at least 3.10 available in their repos. A user would need to be running RHEL 6 or older (which don't even support cmake 3.0) to not have CUDAToolkit support in their cmake install.

This pr essentially makes the requirement cmake 3.10 instead of cmake 3.0.

biochem-fan:
Isn't it just adding if-else with CMake version checks?

FilipeMaia:
@Qiangong2, as @biochem-fan said, I would suggest suggest:
```
IF(CMAKE_VERSION VERSION_LESS 3.10)
FIND_PACKAGE(CUDA)
ELSE(CMAKE_VERSION)
# newer version of CMake use a different module name
FIND_PACKAGE(CUDAToolkit)
ENDIF(CMAKE_VERSION)
```

Qiangong2:
Added in the if/else statement as requested. I can verify it builds and runs without issue.

biochem-fan:
@Qiangong2 @FilipeMaia 

Thank you very much.

> I can verify it builds and runs without issue

Did you test on versions both before and after 3.10?
If so, I will merge this.

Qiangong2:
I don't have a machine old enough to test if it works before 3.10, but since the if statement would use FIND_PACKAGE(CUDA) if it returned true, I believe it is functional.

biochem-fan:
In `FindCUDA`, we can control arguments to `nvcc` via the `CUDA_NVCC_FLAGS` variable.
What is the new way to do this? We need to set the C++ standard and enable OpenMP in `nvcc`.
Failure to do so resulted in [silent but serious problems](https://github.com/3dem/relion/commit/dcc8fe73088aca092d3d4738c3ce79f4731c1bf3) before.

I looked at [FindCUDA documentation](https://cmake.org/cmake/help/latest/module/FindCUDA.html
):

> It is no longer necessary to use this module or call find_package(CUDA) for compiling CUDA code. Instead, list CUDA among the languages named in the top-level call to the [project()](https://cmake.org/cmake/help/latest/command/project.html#command:project) command, or call the [enable_language()](https://cmake.org/cmake/help/latest/command/enable_language.html#command:enable_language) command with CUDA. Then one can add CUDA (.cu) sources directly to targets similar to other languages.

None of the pages linked from here mention `CUDA_NVCC_FLAGS`. Is the variable no longer supported after `FindCUDA` removal? What is the alternative way?

biochem-fan:
I am working on this in #1016. I close this PR.

The required change is more complicated than simply replacing `FIND_PACKAGE(CUDA)`.

PR #1121, opened 2024-05-08
Initialize pre_exposure_micrographs for glob-style input

fullerjamesr:
When provided a glob of movie filenames instead of a .star file as the input --i parameter, initialize the `pre_exposure_micrographs` vector to 0.0 for all input movies.

Absent this, relion_run_motioncorr in ver5.0 with glob input segfaults at motioncorr_runner.cpp line 518 when attempting to recall per-movie additional pre-exposures.

### DIFF
--- a/src/motioncorr_runner.cpp
+++ b/src/motioncorr_runner.cpp
@@ -288,7 +288,8 @@ void MotioncorrRunner::initialise()
 	else
 	{
 		fn_in.globFiles(fn_micrographs);
-		optics_group_micrographs.resize(fn_in.size(), 1);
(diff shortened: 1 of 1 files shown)


biochem-fan:
Thank you very much for your feedback, but I am not sure if we want to keep support of glob inputs in this command. Such a mode of operation seems to be a legacy of an old code path. Our preferred way of input is a STAR file.

@scheres: is non-STAR input used in the STA pipeline?

scheres:
No, the glob input is never used. Better to define things properly in star files than initialising to an arbitrary number.

PR #1143, opened 2024-06-12
Update movie_reconstruct.cpp so movie reconstruct now builds

jdickerson95:
Fixed out of scope error introduced in commit 9d0df80 so relion_movie_reconstruct now builds.


biochem-fan:
No response but closing this because https://github.com/3dem/relion/commit/864ff11231da345c02ab0fac7454e529e6c77a0a has been merged.

Issue #1167, opened 2024-07-23
Relion5 Ctf estimation：  BUG: MDin should contain a rlnMicrographPreExposure label 


kexinliu0001:
hi! I'm doing CtfFind using Relion5.
I did Import and MotionCorr as the tutorial, but when I'm doing CTF estimation with input tilt series corrected_tilt_series.star, it gave the error information "BUG: MDin should contain a rlnMicrographPreExposure label":

![image](https://github.com/user-attachments/assets/dbfbbf47-c60c-4ce9-ac1c-18c0ebf09331)

I checked the star files, and the star file for each tomogram contains the rlnMicrographPreExposure

![image](https://github.com/user-attachments/assets/60b6eedb-16fd-4c0e-8898-2e1d5ae9af29)

How can I solve it?

Thanks!!!

2019C:
Hello, I am running into the same error message in RELION 5 tomo CTF estiomation job. Have you figured out the problem?

BUG: MDin should contain a rlnMicrographPreExposure label

Thanks,
Chen

2019C:
Hi, 

My problem was fixed. It turned out that the CTFFIND4 couldn't run on the cluster. The solution is to add "export LC_ALL=C"  to my ~/.bashrc.

Hope it helps. 
Chen

Li19970731:
I have  added "export LC_ALL=C" to my ~/.bashrc, but this bug still here.

xingxingjiukunkun:
"I have the same issue, have you resolved it?

AssmannG:
HI,
 I had the same issue and could also resolve it with exporting  LC_ALL=C inside the relion SLURM submission templates ...seems as if the $LANG setting that was set on the cluster was not not the standard  "en_US.UTF-8". Accoringly, CtfFind failed.... you can check that with `$ locale ` from where you are running the jobs...

HDuyvesteyn:
Hiya,
I am also encountering the same issue, even after trying exporting LC_ALL=C in my slurm submission script with relion-5.0-beta-3 and ctffind-4.1.14.

MXXXZ:
Thanks @HDuyvesteyn , I changed the ctffind version from 4.1.14 to 4.1.5 and the bug disappeared, so it is indeed likely a bug in the new version.

huwjenkins:
[...]

PR #1212, opened 2024-11-27
ctffind_runner better error message

huwjenkins:
Possible fix for issues like #1167? When CTFFIND fails to fit CTF parameters for all input micrographs (i.e. `MDctf` is empty) exit with an error rather than failing later.


biochem-fan:
@huwjenkins Thank you very much.
Can you also add `LC_ALL=C` to the CTFFIND caller script written by this wrapper?

biochem-fan:
> The user can choose this via RELION_SHELL environment variable

Oh, that is a pain.

Does `env LC_ALL=C ctffind ...` work?

biochem-fan:
Thank you very much for your contribution. I merged your patches.

PR #1258, opened 2025-04-03
CPU kernel patch for oneAPI compiler performance workaround and code refactoring

do-jason:
Add macro function to replace inline function (TRANSLATE_PIXEL_2D, TRANSLATE_PIXEL_3D)
Add "USE_SINCOS_TABLE" macro to CPU acceleration code for alternative code path
Add "#pragma forceinline" for project2Dmodel/project3Dmodel functions
Add alignas in CPU kernel function's arrays
Add "#if _OPENMP" to protect "#pragma omp simd" pragmas from old compiler
Add "#pragma omp simd simdlen" for some cases
Some minor code refactoring
Remove "__attribute__((always_inline))" since "inline" is enough
Remove DEBUG_CUDA
Remove unused no_tex2D and no_tex3D in acc/cpu code path
Remove commented old code block in diff2.h

Compilation checked with GNU compiler 7~13, Intel Classic Compiler, and various Intel oneAPI compiler versions.


biochem-fan:
@do-jason Thank you very much!

Regarding `Add --enable-avx2 --enable-avx512 --enable-fma for FFTW build`: what happens with consumer CPUs without AVX512? Will the binary run fine without using AVX512 or cause illegal instruction errors?

FFTW documentation says:

> Enable various SIMD instruction sets. You need compiler that supports the given SIMD extensions, but FFTW will try to detect at runtime whether the CPU supports these extensions. 
>  That is, you can compile with--enable-avx and the code will still run on a CPU without AVX support.
https://www.fftw.org/fftw3_doc/Installation-on-Unix.html

So it seems safe but am I correct?

biochem-fan:
With this patch, is the performance of oneAPI compilers approaching Classic Intel compiler?

do-jason:
> @do-jason Thank you very much!
> 
> Regarding `Add --enable-avx2 --enable-avx512 --enable-fma for FFTW build`: what happens with consumer CPUs without AVX512? Will the binary run fine without using AVX512 or cause illegal instruction errors?
> 
> FFTW documentation says:
> 
> > Enable various SIMD instruction sets. You need compiler that supports the given SIMD extensions, but FFTW will try to detect at runtime whether the CPU supports these extensions.
> > That is, you can compile with--enable-avx and the code will still run on a CPU without AVX support.
> > https://www.fftw.org/fftw3_doc/Installation-on-Unix.html
> 
> So it seems safe but am I correct?

It will just enable features and it will choose best instruction set at runtime. So, enabling it is safe.

> With this patch, is the performance of oneAPI compilers approaching Classic Intel compiler?

In my experiment, the difference became single digit percent Plasmodium Ribosome workload. There could be compiler improvement, but I cannot guess when it will be possible. This is workaround for the time being.

biochem-fan:
Thanks for clarification.

> Compilation checked with GNU compiler 7~13, Intel Classic Compiler, and various Intel oneAPI compiler versions.

Trusting this, I will merge this into the `ver5.0` branch. If nobody reports errors, I will merge this to `master` and include in the 5.0.1 minor update.

Thanks again for your contribution.

PR #1267, opened 2025-05-15
parallel applyHelicaSymmetry

estrozi:
Dear Developers,

I have make some changes to the applyHelicalSymmetry function (backprojector.cpp) to make it faster with the use of threads.

In many helical projects, it's common to evaluate numerous (>50) different symmetry parameters through full 3D auto-refinement jobs. This scenario becomes particularly computationally demanding for structures characterized by very small axial rise values. A notable example is EMD-42436 (https://www.ebi.ac.uk/emdb/EMD-42436), which has an axial rise of just 0.09892 Å. Such structures typically have over 500 unique helical asymmetric units per segment box, making the symmetry imposition step a significant portion of the total processing time.

With these modifications, I was able to reduce the processing time for auto-refinement from approximately 7 hours to 4.5 hours (for the same number of iterations).

I hope this (made into a separated branch of mine) could be integrated to the main project alongside the previously discussed (far from perfect but still useful) pull request (https://github.com/3dem/relion/pull/1185) which adds functionality for calculating the average helical PS in the displayer.

Best wishes,

Leandro F. Estrozi.

### DIFF
--- a/src/backprojector.cpp
+++ b/src/backprojector.cpp
@@ -2140,7 +2140,7 @@ void BackProjector::symmetrise(int nr_helical_asu, RFLOAT helical_twist, RFLOAT 
 	enforceHermitianSymmetry();
 
 	// Then apply helical and point group symmetry (order irrelevant?)
-	applyHelicalSymmetry(nr_helical_asu, helical_twist, helical_rise);
+	applyHelicalSymmetry(nr_helical_asu, helical_twist, helical_rise, threads);
 
 	applyPointGroupSymmetry(threads);
 }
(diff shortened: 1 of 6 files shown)


estrozi:
Hi ! The rotations being exclusively are around Z the threads could be made z-slice independent with 'small' 2D buffers. So now it goes faster and it does not have a significant impact on the overall memory consumption. I hope this becomes acceptable.

Best wishes,

Leandro.

biochem-fan:
@estrozi Thank you very much. Your 2D local array is a very clever trick :)

@scheres I think this is a nice contribution. Because this touches the core code, I suggest merging this into the private 5.1 branch under development to allow more testing at LMB. What do you think?

estrozi:
> ok, great! I am adding this code to the master branch of our private relion-devel. It will then become available to the public later this year, or the beginning of next. S

Thanks a lot ! I would also really appreciate if the other PR about the PS average could be incorporated. But if it's not possible, that's fine.

PR #1271, opened 2025-06-06
Update eer_trajectory_handler.py (fix NumPy deprecation of np.int and np.float in RELION 5 conda)

rbs-sci:
Update eer_trajectory_handler.py to fix NumPy deprecation of np.int and np.float in RELION 5.0 conda environment.


biochem-fan:
Great! Thank you very much for your contribution.

PR #1283, opened 2025-07-28
parallel applyHelicalSymmetry and PS averaging

estrozi:
Dear developers,

I've been working hard to create a PR where:

- I modified the result produced by relion_image_handler with the option "--avg_ampl2_ali" so it writes a full (not only half) PS, well centered, padded with a padding factor and log-normalized.
 - I have added the option "--PS_padding_factor" (default: 2).
- I have added the option "--avg_phase_ali" to produce the equivalent of "--avg_ampl2_ali" for the PS phases.
- I replaced my previous non-elegant use of displayer.cpp to average the PS by a more simple call to relion_image_handler followed by a call to relion_display.

I hope this time I achieved the simplicity/efficiency that is necessary for this PR to be accepted.

As always, thanks for all your efforts, best wishes,

Leandro.

PS: Don't worry I won't give up ;-).
PS2: Below just to illustrate (one more time) the difference between the PS (amplitudes/phases) of a 2D class and the average of class members PS (amplitudes/phases) for a TMV.

best_2Dclass:
best_2Dclass
amplitudes_of_best_2Dclass:
amplitudes_of_best_2Dclass
phases_of_best_2Dclass:
phases_of_best_2Dclass
average_amplitudes_of_members_of_best_2Dclass:
average_amplitudes_of_members_of_best_2Dclass
average_phases_of_members_of_best_2Dclass:
average_phases_of_members_of_best_2Dclass


estrozi:
:-(

Issue #1280, opened 2025-07-23
Relion 5 subtomogram extraction tomogram centering discrepancy


daniel-ji:
See the README.md here: https://github.com/daniel-ji/relion-subtomo-rounding-bug. Let me know if there's any other questions / comments / concerns! Looping in @uermel as well. Thanks!

biochem-fan:
This is too STA-specific for me to comment; as far as I understand, RELION assumes all image/particle dimensions to be even numbers. I don't know why they set -999 as the default and why they permit cases where the input dimension is not specified in a STAR file.

@scheres who is in charge of these codes?

PR #1286, opened 2025-07-29
fix: use floating point division for subtomogram extraction projection matrix

daniel-ji:
Addresses #1280. See issue description for more details regarding the fix. In short, this fix ensures that there's no 0.5px error in the extracted subtomograms when tomogram dimensions are odd / not provided.

Thanks!

Best,
Daniel


biochem-fan:
OK, sorry for my confusion. I merged #1285.

This one is trickier and I cannot judge. In SPA, we intentionally use integer division to round down. I don't know the intended behavior here.

Issue #1141, opened 2024-06-07
Tilt angle offset being undone when aligning TS with IMOD


rdrighetto:
Hi,

I noticed that when aligning TS using patch-tracking in IMOD from RELION-5, a tilt angle offset is estimated in the "Fine Alignment" step equivalent of Etomo:
![image](https://github.com/3dem/relion/assets/7696255/ca104f63-59c1-44d2-9268-eb15e911770f)
I didn't even know this was possible, but I like it. If i proceed with the tomogram reconstruction in Etomo I get a nicely flattened lamella in the XZ plane:
![image](https://github.com/3dem/relion/assets/7696255/72dc2929-3528-4d0a-9b38-056d3ab23ecd)

However, upon proceeding to the "Reconstruct tomograms" step in RELION, I see that the lamella is tilted as if the tilt angle offset had not been applied:
![image](https://github.com/3dem/relion/assets/7696255/80fd8be0-a0ea-465d-a713-2891eb9dc0de)
This is confirmed by inspecting the `_rlnTomoYTilt` field in the respective tilt series STAR file.

Upon inspecting the code of the align tilt-series job, I found that the tilt angle offset is being undone by these lines:
https://github.com/3dem/relion/blob/6331fe600cca7683ecf7c1011ce676701faf1e97/src/tomography_python_programs/align_tilt_series/imod/_utils.py#L28C1-L29C41

Is this a bug, or is it meant to be this way? Or did the IMOD behavior in this regard has changed across versions?

**Environment:**
 - OS: Ubuntu 22.04.3 LTS
 - RELION version: RELION-5.0-beta-3-commit-63311fe

**Job options:**
 - Type of job: Align tilt-series
 - IMOD version: 4.12.61

Thank you!

ShelvinZhang:
Hi Ricardo,

Have they answered your question, or do you already have an answer to this?

I realized the same issue when I tried to do Alignment using IMOD in Relion5. The output star file shows identical values for _rlnTomoYTilt and _rlnTomoNominalStageTiltAngle. From my understanding, there should be an offset between these two parameters, and they do differ if Aretomo was used instead of IMOD.

I noticed that the tilt parameters are all gone after tomogram reconstruction. I'm a complete rookie in cryoET, so I don't know if this could cause any problem to the following process. Do you have any insights into this?

Best,
Shuwen

ShelvinZhang:
Well, this issue was explained in the Relion5 paper:

[...]

PR #1343, opened 2026-05-16
Preserve IMOD final tilt angles in YTilt

amineuron:
## Summary
- keep IMOD final `.tlt` angles in `_rlnTomoYTilt` instead of subtracting `AngleOffset`
- make the C++ align_tiltseries runner and Python IMOD helper consistent
- document that `.tlt` is IMOD's final OutputTiltFile used for reconstruction geometry

## Rationale
When IMOD `tiltalign` estimates `AngleOffset`, the final `.tlt` file already includes that offset and IMOD `tilt` reconstructs from those corrected angles. Subtracting `AngleOffset` while writing `_rlnTomoYTilt` reverts the STAR metadata to nominal stage tilts, making RELION projection matrices inconsistent with the IMOD reconstruction geometry. This matches the symptom reported in #1141.

## CTF / downstream extraction
This PR intentionally does not rewrite existing `_rlnCtfScalefactor` values during alignment. RELION native tomogram reconstruction already initializes `_rlnCtfScalefactor = cos(_rlnTomoYTilt)` when the column is absent, so preserving the final IMOD angles keeps the default CTF scale consistent with the projection geometry.

If `_rlnCtfScalefactor` is already present, it may have been written by RELION CTF refinement as a signal/ice-thickness scale. Overwriting it at alignment-import time would be unsafe.

## Tests
- `git diff --check`
- `python -m py_compile src/tomography_python_programs/align_tilt_series/imod/_utils.py`

### DIFF
--- a/src/align_tiltseries_runner.cpp
+++ b/src/align_tiltseries_runner.cpp
@@ -658,8 +658,8 @@ bool AlignTiltseriesRunner::readIMODResults(long idx_tomo, std::string &error_me
     std::string line;
     std::vector<std::string> words;
 
-    // 1. Get tiltangle_offset from the align.log file
-    RFLOAT tiltangle_offset = 0., mean_error = 0., stddev_error = 0., leaveout_error = 0.;
+    // 1. Get alignment diagnostics from the align.log file
+    RFLOAT mean_error = 0., stddev_error = 0., leaveout_error = 0.;
     std::ifstream in0(fn_align.data(), std::ios_base::in);
     if (in0.fail())
     {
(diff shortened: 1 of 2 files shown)


amineuron:
Closing this PR.

After deeper investigation in my fork, I could see you might had designer choice and might need to be sure this convention change would be safe, couldn't find why Relion choose different convention  


1. **Whether `_rlnTomoYTilt` stores IMOD's pitch-corrected angle or the nominal angle is empirically equivalent for self-consistent project chains** — every YTilt consumer in the built tomo code (`Tomogram::getCtf`, `reconstruct_tomogram.cpp`, `subtomo.cpp`, `reconstruct_particle.cpp`) uses whatever YTilt is in the STAR. 

2. For our own fork the IMOD-consistent convention is convenient, but I don't think it's the right thing to push upstream, it would change default behavior for every existing user without solving a reproducible bug.

Will submit a focused PR for the `_rlnTomoVisibleFrames` issue (`reconstruct_particle.cpp` uses geometric visibility only, while `exp_model.cpp` correctly intersects with the explicit mask, so Reconstruct and Refine3D end up using different frame sets when `--max_dose` was set at Extract).

Thanks for the review patience.

PR #1346, opened 2026-05-30
Treat ".gain" gain references as TIFF on the generic image path

amineuron:
### Problem

A gain reference that is a TIFF but carries a non-TIFF extension — for example the `.gain` files written alongside some Falcon/EER datasets — fails in MotionCorr (own implementation) and in Bayesian Polishing when the movies are **non-EER (TIFF)**, with:

```
ERROR:
Invalid Spider file:  XXXX_GainReference.gain
in: src/rwSPIDER.h, line 151
```

The format is chosen from the file extension alone (`fImageHandler::openFile`), so an unrecognised extension falls through to the SPIDER reader in `Image::_read()`. The EER path already handles this case — `EERRenderer::loadEERGain()` appends `:tif` for `.gain` — and the GUI lists `.gain` as a valid gain extension (`*.{mrc,gain}`), so the file is expected to work; it just doesn't on the generic image path used for TIFF movies.

### Fix

When the extension is not a recognised image format, sniff the first four bytes and read the file as TIFF if it carries the TIFF magic (`II*\0` little-endian or `MM\0*` big-endian). Files with a known extension are untouched, and a non-TIFF unknown extension still falls back to SPIDER exactly as before — the change only ever *promotes* an otherwise-unhandled file to TIFF when the bytes confirm it.

This makes a TIFF gain read consistently on the generic image path, in the same orientation as the (readTIFF) movie frames, so the multiplicative gain correction stays pixel-aligned. The EER-only Y-flip reversal in `loadEERGain()` is intentionally not duplicated here: it exists to reconcile readTIFF orientation with the EER renderer's frame orientation, whereas TIFF movie frames already go through readTIFF, so no reversal is needed.

### Scope

- Restricted to read-only opens of existing files with an unrecognised extension.
- Fixes the shared `Image::read` path, so both MotionCorr and Bayesian Polishing pick it up.
- The EER path is unaffected (its `:tif` override means the extension is already recognised when `Image::openFile` runs).

### DIFF
--- a/src/image.h
+++ b/src/image.h
@@ -302,6 +302,16 @@ public:
 			fileName = fileName.addExtension(ext_name);
 		}
 
+		// A ".gain" reference is a TIFF. RELION already assumes this in
+		// EERRenderer::loadEERGain (which appends ":tif" for .gain); make the
+		// generic image path assume the same, so a TIFF gain with a .gain
+		// extension also reads for non-EER (e.g. TIFF) movies and in Bayesian
+		// Polishing, instead of falling through to the SPIDER reader.
+		// See https://github.com/3dem/relion/pull/1346 and the CCPEM thread:
+		// https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind2605&L=CCPEM&O=D&P=34021
+		if (ext_name == "gain")
+			ext_name = "tif";
+
 		isTiff = ext_name.contains("tif");
 
 		// Open image file


biochem-fan:
Related: https://www.jiscmail.ac.uk/cgi-bin/wa-jisc.exe?A2=ind2605&L=CCPEM&O=D&P=34021

biochem-fan:
Thank you very much for this.

Other possibilities would:
- Do this magic byte check only for `.gain` files
- Always treat `.gain` as TIFF without checking the magic byte

Do you see problems with these alternatives?

I don't expect people use the SPIDER format for a gain reference in the foreseeable future.

amineuron:
Thanks! Good point — I have switched to option (2): treat `.gain` as TIFF by extension, which is exactly what `EERRenderer::loadEERGain` already does (it appends `:tif` for `.gain` before reading). So this is not a new assumption; it just makes the generic `Image::read()` path consistent with the EER path.

That asymmetry is also why the original report only showed up with non-EER (TIFF) movies: MotionCorr routes the gain through `loadEERGain` for EER movies (which handles `.gain`) but through `Image::read` for TIFF movies (which did not). Keeping the fix in `Image::read` means Bayesian Polishing and the other non-EER gain readers get the same handling. I dropped the magic-byte check.

amineuron:
I went with the checking the byte cause I didn't want to assume .gain is always TIFF, sniffing the bytes felt more general/robust. But loadEERGain already makes that assumption, so it's simpler to just mirror it than add a separate detector.

biochem-fan:
The original reason I didn't blindly assume `.gain` == `.tiff` was that TFS generated some gain files in 4K and others in 8K or even 16K (in early days of EER). Because EER could be rendered in 4, 8 or 16K, I had to implement gain resizing. My idea was to handle the gain reference in a special code path when reading EER files, while `relion_convert_to_tiff` writing an MRC gain reference in the correct size when converting EER to TIFF. Now that TFS started to write TIFF files with a `.gain` gain reference, the situation got complicated. I *hope* their `.gain` file always matches the size of TIFF movies.

amineuron:
Agreed. The super-resolution/resizing case really belongs to EER, and if someone ever feeds a mismatched gain, it's not silent MotionCorr will stop with the "gain size doesn't match" error. So this just covers the direct TIFF-movie + .gain case, while EER keeps its own resizing logic.

biochem-fan:
OK. I agree to merge your fix. Could you please add the URLs of this PR and the CCPEM thread to the code comment?

amineuron:
Done — added both the PR link and the CCPEM thread URL to the code comment. Thanks Takanori!

biochem-fan:
I merged this. Thank you very much for your contribution.

PR #1354, opened 2026-08-07
Fix uninitialized MPI count read for queued jobs

stanbot8:
Queued job preparation can leave `nr_mpi` uninitialized because it is populated while constructing local commands. The final guard evaluated `nr_mpi` before checking `do_queue`, causing queued jobs to read it unnecessarily. Checking `do_queue` first short-circuits that read while preserving the existing MPI warning for local jobs.


biochem-fan:
@stanbot8 Thanks. This makes sense. I merged it.

PR #1359, opened 2026-08-14
fix: make patch motion groups the sum of the underlying frame data

fullerjamesr:
I noticed while playing around with motion correcting movies with intentionally inadvisably high fractionation that grouping was not having the expected result. I think there is a bug here where currently a patch's group actually just ends up set to the contents of the last frame in the group, whereas I assume what I have in this pull request was/is the actual intent: sum the contents of the frames within each group/patch?

When I tested this change, grouping gave better/more successful local motion results than the existing code.


biochem-fan:
Oops! You are right. This was a terrible mistake.

Thank you very much for catching it!

Issue #1319, opened 2025-12-18
MultiBody MPI send/receive buffer size mismatch


fullerjamesr:
**Describe your problem**

There appear to be corner cases during MultiBody refinement in the `MlOptimiserMpi::combineAllWeightedSums()` function where the send/receive buffer size (the calculated size of `Msum`) for one MPI rank becomes mismatched, resulting in an `MPI_ERR_TRUNCATE` error and termination.

This seems to happen under the following conditions:
1. At least one body is being kept fixed; does not have to be all bodies but can be.
2. (This one is a little less clear to me) The healpix sampling is due to be (ought to be?) decreased but then isn't? Or something like that. More on this in the following paragraph/list of observations:

- When I restart a crashed job with `--force_converge`  the healpix sampling in that final iter that it runs is a step finer than the iteration in which the crash happened
- Similarly, when I run an identical job with `Combine iterations through disc`, the final to-nyquist iteration is most often the same iteration number as the crashed one, but again uses a finer sampling just for that final iteration
- I've found that an easy way for me to test this (trigger the crash or not) is to change whether the solvent is filled with zeros or noise, the latter of which increases (worsens) angular accuracy values. The issue seems to occur when the solvent is noise and thus the angular accuracy is a bit iffy, but still should decrease (i.e., the angular accuracy is outside the 90% threshold for at least one body but potentially all bodies, causing it to be kept fixed, but within the 75% threshold for stepping down the healpix sampling). If I start a duplicate job starting from a finer healpix sampling to begin with so that all bodies are instead totally outside the 75% accuracy threshold once the run approaches convergence, everything proceeds normally and the run completes.

The easy workaround is of course to `Combine iterations through disc` which works (albeit a bit slower with all that disc IO, which can be challenging for large data depending on the storage setup). But I thought I should report this in case it's helpful.

To try to be helpful I enabled the `DEBUG` blocks in `MlOptimiserMpi::combineAllWeightedSums()` and ran a job that crashes. I will attach the full `run.out` and `run.err` files for that here. You can see what I think (?) is the mismatched MPI buffer size a few lines before the errors start getting printed:

` AA SEND node->rank= 1 MULTIDIM_SIZE(Msum)= 29094236 this_follower= 1 other_follower= 3`

...it looks like rank 1 is calculating a slightly too-small size for `Msum` compared to the other ranks.

Also attached is the `run.out` from the identical job but using `Combine iterations through disc` that finishes successfully.

**Environment:**
[...]

PR #1362, opened 2026-08-24
Fix MPI buffer-size mismatch in multibody combineAllWeightedSums (MPI_ERR_TRUNCATE)

fullerjamesr:
Found the fix for my issue #1319  

In `MlOptimiserMpi::calculateExpectedAngularErrors()`, MPI rank 1 ("first_follower") calculates the per class/body rotation and translation accuracies. It syncs the best (minimum numerical value) of both vectors (stored in the scalar `acc_trans` and `acc_rot`) to the other MPI processes, who then each run the logic in `updateAngularSampling()` themselves. But as they're all operating from synced data they should reach the same conclusions, and most decisions are done based on these single "best" values so everything is fine.

Except that the decision in `updateAngularSampling()` whether or not to hold a particular body fixed needs the individual per-body values, thus the full vector. As it stands, this is not broadcast/synced and so the fixed-ness of bodies can diverge on other ranks because they never change their copy of this vector from the default initialization of all zeros. When bodies are being fixed on rank 1, this can lead to different `wsum_model.pdf_direction[]` sizes versus the other ranks and the `MPI_ERR_TRUNCATE` crash I describe in that issue.

This adds Bcast of the full per-body/class accuracies alongside the existing best/minimum value Bcast.

(A almost-entirely-pointless side effect of this is that, even for normal Refine3D runs, `run_it*_half2_model.star` files will now have real values instead of zeros in the printed `data_model_classes->rlnAccuracyRotations` & `rlnAccuracyTranslationsAngst` fields, whereas currently those are all zeros)

### DIFF
--- a/src/ml_optimiser_mpi.cpp
+++ b/src/ml_optimiser_mpi.cpp
@@ -3661,6 +3661,9 @@ void MlOptimiserMpi::calculateExpectedAngularErrors(long int my_first_part_id, l
 	// The reconstructing follower Bcast acc_rottilt, acc_psi, acc_trans to all other nodes!
 	node->relion_MPI_Bcast(&acc_rot, 1, MY_MPI_DOUBLE, first_follower, MPI_COMM_WORLD);
 	node->relion_MPI_Bcast(&acc_trans, 1, MY_MPI_DOUBLE, first_follower, MPI_COMM_WORLD);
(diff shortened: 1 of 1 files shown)


biochem-fan:
Wow! This is a great discovery. Thank you very much.

The change looks OK to me but I am not very confident about the internal details of `MlOptimiser`.

@scheres Could you please confirm this?

PR #1369, opened 2026-09-16
Reuse exactly-sized free blocks in the custom GPU allocator (#1225)

### LINKED ISSUE #1225: There is a bug in custom_allocator.cuh
**location：**
_getFirstSuitedFree function in custom_allocator.cuh
**reason**
The size of the first suited allocation can be equal to the target size.  The function  Alloc* alloc(size_t requestedSize)  also proves this point
**modification**
while (a != NULL && ( a->size < size || ! a->free ) )

arcusbuilds:
Fixes #1225, reported by @fengxiaoshuai.

`_getFirstSuitedFree()` skips a free block whose size is *exactly* the requested size:

```cpp
while (a != NULL && ( a->size <= size || ! a->free ) )
```

That makes the `curAlloc->size == size` branch in `alloc()` unreachable, and a request the pool can satisfy exactly instead aborts with `CRITICAL(ERRGPUCAOOM)`. It also stops a freed block from being reused by a request of the same size, which fragments the pool. Since requests are rounded up to `alignmentSize` first, exact fits are common.

Changed `<=` to `<`. The HIP port has an identical copy of the function, so both are updated.

Not compiled against CUDA/HIP here (the build workflow is CPU-only, so it does not cover these headers) — checked by running the free-list logic standalone with the CUDA calls stubbed out.


biochem-fan:
Thank you very much. I merged this into the `ver5.1` (beta version) branch to let others test the code.

arcusbuilds:
@biochem-fan hey, quick one, is there any CUDA related work coming up in the repo? wanted to see if there's something in that area I could jump in on.

biochem-fan:
@arcusbuilds we have already did a big refactoring of the GPU code path but it is still being tested internally and not public yet.

